PR #3, opened 2021-12-07
COP-9029: Show and hide logic for components

paulfgraham:
### Description
Mechanism for determining whether a field should be shown or not, based on the value of one or more other fields.
https://support.cop.homeoffice.gov.uk/browse/COP-9029

### To test
This should be covered by the unit tests within this commit. For the benefit of the QA team, this will be demonstrable within the Storybook stories in a subsequent PR.


jennikate (approved):
LGTM

PR #4, opened 2021-12-07
COP-9575: Add mechanism for getting component based on configuration

paulfgraham:
### Description
The configuration dictates what sort of component to get and the component returned is the one from the `cop-react-components` library.

In addition to this, the `radios` and `autocomplete` types require options, which necessitates functions for obtaining those.

The logic for retrieving reference data by URL has been mocked within the library for the time being - a proper implementation that uses the URL properly will be needed but that is out of scope for this ticket and PR.

Related JIRA ticket: https://support.cop.homeoffice.gov.uk/browse/COP-9575.
(Also part of https://support.cop.homeoffice.gov.uk/browse/COP-8975)

### To test
This should all be covered by the unit tests. For the benefit of QAs, Storybook stories with appropriate JSON configuration will be in a later PR.


michaelruocco (approved):
LGTM

PR #6, opened 2021-12-16
COP-9615: Load reference data options

paulfgraham:
### Description
Added a mechanism for loading reference data. This involves a series of hooks and an interceptor pattern to allow auth tokens to be added to requests (for example).
https://support.cop.homeoffice.gov.uk/browse/COP-9615

### To test
The are extensive unit tests around all of the new functionality.


src/utils/Component/cleanAttributes.test.js
```diff
@@ -16,11 +16,11 @@ describe('utils', () => {
       });
 
       it('should handle empty options', () => {
-        expect(cleanAttributes({})).toEqual({});
+        expect(cleanAttributes({}, null)).toEqual({});
       });
 
       it('should handle a non-object options', () => {
-        expect(cleanAttributes(23)).toEqual({});
+        expect(cleanAttributes(23, null)).toEqual({});
```
michaelruocco, line 23: what is significant about the value 23 here? or is it just an arbitrary not-object option?

src/utils/Component/getComponent.js
```diff
@@ -73,9 +73,9 @@ const getComponentByType = (config) => {
   }
 };
 
-const getComponent = (config) => {
+const getComponent = (config, noWrap = false) => {
   const component = getComponentByType(config);
-  if (component && isEditable(config)) {
+  if (component && !noWrap && isEditable(config)) {
```
michaelruocco, line 78: I didn't noticed this when we discussed before but I often find double negatives like "not no wrap" a bit confusing, could this be made to be called `wrap` and then defaulted to true? so that you could do `!wrap` instead, or does that have knock on effects elsewhere?

src/hooks/useGetRequest.js
```diff
@@ -0,0 +1,68 @@
+// Global imports
+import axios from 'axios';
+import { useEffect, useState } from 'react';
+
+// Local imports
+import useAxios from './useAxios';
+
+const cache = {};
```
michaelruocco, line 8: I suspect this could be something to deal with in future rather than now, but Ill ask anyway, does this cache ever get cleared? or will it just keep growing forever? If only a handful of requests are going to be made and a new instance is used for each component then its probably find, but if one instance of the cache is reused across all components I suppose it might get quite large?

michaelruocco (commented):
I've raised a few comments that I am not entirely sure make sense, but I figured I'd ask them anyway!

src/utils/Component/getComponent.js
paulfgraham: That's a fair point. Happy to change it.

src/hooks/useGetRequest.js
paulfgraham: Done.

src/utils/Component/getComponent.js
paulfgraham: Done.

PR #11, opened 2021-12-22
COP-9577: Add PageActions component

paulfgraham:
### Description
Added the `PageActions` and accompanying `ActionButton` components, along with unit tests.

These are displayed at the bottom of a page, based on the `"actions"` property within the JSON.

### To test
Both components in this PR have 100% unit test coverage.

### DIFF
--- a/src/components/PageActions/ActionButton.jsx
+++ b/src/components/PageActions/ActionButton.jsx
@@ -17,7 +17,7 @@ const ActionButton = ({
     return null;
   }
   return (
-    <Button {...attrs} className={action.className} onClick={() => onAction(action)}>
+    <Button {...attrs} className={action.className} classModifiers={action.classModifiers} onClick={() => onAction(action)}>
       {action.label || DEFAULT_LABEL}
     </Button>
   );
--- a/src/components/PageActions/ActionButton.jsx
+++ b/src/components/PageActions/ActionButton.jsx
@@ -1,0 +1,31 @@
+// Global imports
+import { Button } from '@ukhomeoffice/cop-react-components';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export const DEFAULT_LABEL = 'Continue';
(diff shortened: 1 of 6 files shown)


src/components/PageActions/ActionButton.test.js
```diff
@@ -0,0 +1,68 @@
...
+    const ON_ACTION = (action) => {
+      ON_ACTION_CALLS.push(action);
+    };
+    
+    beforeEach(() => {
+      ON_ACTION_CALLS.length = 0;
+    });
+
+    it('should handle a null action', async () => {
+      const ACTION = null;
```
Ali-Baker, line 21: Test throw warning here because of `null`

```
 Warning: Failed prop type: The prop `action` is marked as required in `ActionButton`, but its value is `null`.
          at ActionButton (/Users/aliali/code/cop/cop-react-form-renderer/src/components/PageActions/ActionButton.jsx:11:11)
```

src/components/PageActions/ActionButton.jsx
```diff
@@ -0,0 +1,31 @@
...
+  action: _action,
+  onAction,
+  ...attrs
+}) => {
+  const action = typeof(_action) === 'string' ? DEFAULT_ACTIONS[_action] : _action;
+  if (!action) {
+    return null;
+  }
+  return (
+    <Button {...attrs} className={action.className} onClick={() => onAction(action)}>
```
Ali-Baker, line 20: If no `className` exists then `class="undefined"`

Why don't you give it a default value of empty string

src/components/PageActions/ActionButton.test.js
paulfgraham: Good spot. That wasn't showing up as a lint warning and I missed it in the unit test run. Have removed the test as it's redundant.

Ali-Baker (approved):
🚢 it

PR #28, opened 2022-01-04
COP-9700: Add canActionProceed helper

paulfgraham:
### Description
Added the `canActionProceed` helper function to determine whether an action invoked on a form page is able to proceed. This checks whether the page should be validated and, if so, validates it and handles the outcome appropriately.

### To test
This should be covered by the accompanying unit tests.


jgar123 (approved):
LGTM

PR #41, opened 2022-01-10
Change JSON data structure

paulfgraham:
### Description
The initial data structure has changed to remove the various contexts and replace them with something cleaner. This was specifically requested following the initial deployment last week.

### To test
If all of the Storybook stories still work then this has succeeded. There are no coding changes.


Ali-Baker (approved):
🚀

PR #51, opened 2022-02-07
COP-10048: Add TextArea component

paulfgraham:
### Description
Add support for `TextArea` components, which is now available in the COP React Components library.

### To test
A unit test has been added to make sure the component is appropriately rendered from a JSON snippet. In addition, it's testable within the Storybook Sandbox.


src/utils/Component/getComponent.test.js
```diff
@@ -262,6 +262,40 @@ describe('utils', () => {
...
+        rows: ROWS,
+        'data-testid': ID
+      };
+      const { container } = render(getComponent(COMPONENT));
+
+      const [ formGroup, textarea ] = getAllByTestId(container, ID);
+      expect(formGroup.tagName).toEqual('DIV');
+      expect(formGroup.classList).toContain('govuk-form-group');
+      let label = undefined;
+      formGroup.childNodes.forEach(node => {
```
gardnersam, line 284: Wouldn't we expect the Label to be the 0th childNode? Followed by hint and error as positions 1 and 2

paulfgraham: Yes, I suspect so. I simply copied the other tests on this one. Happy to change it, though.

gardnersam (approved):
LGTM

src/utils/Component/getComponent.test.js
paulfgraham: Done.

PR #52, opened 2022-02-09
EAB-85 Add date component

gardnersam:
**Description**
Add support for DateInput components, which is now available in the COP React Components library.

**To test**
A unit test has been added to make sure the component is appropriately rendered from a JSON snippet. In addition, it's testable within the Storybook Sandbox.

### DIFF
--- a/src/models/ComponentTypes.js
+++ b/src/models/ComponentTypes.js
@@ -7,6 +7,7 @@ const TYPE_PHONE_NUMBER = 'phone-number';
 const TYPE_RADIOS = 'radios';
 const TYPE_TEXT = 'text';
 const TYPE_TEXT_AREA = 'textarea';
+const TYPE_DATE = 'date';
 
 const ComponentTypes = {
   AUTOCOMPLETE: TYPE_AUTOCOMPLETE,
@@ -17,7 +18,8 @@ const ComponentTypes = {
   PHONE_NUMBER: TYPE_PHONE_NUMBER,
   RADIOS: TYPE_RADIOS,
   TEXT: TYPE_TEXT,
-  TEXT_AREA: TYPE_TEXT_AREA
+  TEXT_AREA: TYPE_TEXT_AREA,
+  DATE: TYPE_DATE,
 };
 
 export default ComponentTypes;
--- a/src/models/ComponentTypes.js
+++ b/src/models/ComponentTypes.js
@@ -1,6 +1,6 @@
 const TYPE_AUTOCOMPLETE = 'autocomplete';
-const TYPE_EMAIL = 'email';
 const TYPE_DATE = 'date';
+const TYPE_EMAIL = 'email';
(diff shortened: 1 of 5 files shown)


src/models/ComponentTypes.js
```diff
@@ -7,6 +7,7 @@ const TYPE_PHONE_NUMBER = 'phone-number';
 const TYPE_RADIOS = 'radios';
 const TYPE_TEXT = 'text';
 const TYPE_TEXT_AREA = 'textarea';
+const TYPE_DATE = 'date';
```
paulfgraham, line 10: Alphabetical...? 😁

src/utils/Component/getComponent.js
```diff
@@ -1,6 +1,6 @@
 // Global imports
 import React from 'react';
-import { Autocomplete, Heading, InsetText, Markup, Radios, TextArea, TextInput } from '@ukhomeoffice/cop-react-components';
+import { Autocomplete, Heading, InsetText, Markup, Radios, TextArea, TextInput, DateInput } from '@ukhomeoffice/cop-react-components';
```
paulfgraham, line 3: Guess what I'm going to say here... 😂

src/utils/Component/getComponent.js
```diff
@@ -56,6 +56,11 @@ const getTextInput = (config) => {
   return <TextInput {...attrs} />;
 };
 
+const getDate = (config) => {
```
paulfgraham, line 59: And again...

src/utils/Component/getComponent.js
```diff
@@ -74,6 +79,8 @@ const getComponentByType = (config) => {
       return getAutocomplete(config);
     case ComponentTypes.RADIOS:
       return getRadios(config);
+    case ComponentTypes.DATE:
```
paulfgraham, line 82: This one is where alphabetical is of no consequence - ideally, we want them ordered according to which will get hit first so we can return as early as possible. Nothing to see or do here as we have no metrics on that right now. I suspect HTML will still be at the top, then probable text inputs, but then who knows?

src/utils/Component/isEditable.js
```diff
@@ -7,7 +7,8 @@ export const EDITABLE_TYPES = [
   ComponentTypes.PHONE_NUMBER,
   ComponentTypes.RADIOS,
   ComponentTypes.TEXT,
-  ComponentTypes.TEXT_AREA
+  ComponentTypes.TEXT_AREA,
+  ComponentTypes.DATE
```
paulfgraham, line 11: D before T, except after 3.

paulfgraham (changes requested):
Just some alphabetisation. 😁

paulfgraham (changes requested):
Close, but no acgir. 🚬

paulfgraham (approved):
Kloos dgoo ot em. 👍

PR #54, opened 2022-02-21
COP-10035: Refactor actions

paulfgraham:
### Description
Refactored page actions and the **Check your answers** links to use `page` as a property rather than mixing and matching `href` and `url`. We presently only ever navigate internally within the form so `page` makes the most sense. This refactor is necessary for saving the form status in a later PR.

https://support.cop.homeoffice.gov.uk/browse/COP-10035

### To test
All of the affected unit tests have been updated, as have the various Storybook stories and JSON files. You can test this by making sure that you can still navigate around within the form in Storybook.

### Release note
This is a major release due to the refactor - existing JSON files (of which there should be almost none in the wild) - will need to be adjusted along the lines of:
```
"pages" : [
  {
    ...
    "actions": [
      { "type": "navigate", "label": "Next", "url": "/page-x" } // Replace this...
      { "type": "navigate", "label": "Next", "page": "page-x" } // ... with this
    ],
    "cya_link": [
      { "url": "/this-page" } // And replace this...
      { "page": "this-page" } // ... with this
    ]
  }
]
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "1.2.1",
+  "version": "2.0.0-alpha",
   "private": false,
   "scripts": {
     "clean": "rimraf dist",
@@ -16,7 +16,7 @@
     "post-compile": "rimraf dist/*.test.* dist/**/*.test.* dist/**/*.stories.* dist/docs dist/json dist/assets"
   },
   "dependencies": {
-    "@ukhomeoffice/cop-react-components": "^0.8.1",
+    "@ukhomeoffice/cop-react-components": "^0.8.2",
     "axios": "^0.21.1",
     "govuk-frontend": "^3.13.0",
     "web-vitals": "^1.0.1"
--- a/src/components/SummaryList/helpers/getRowActionAttributes.js
+++ b/src/components/SummaryList/helpers/getRowActionAttributes.js
@@ -2,8 +2,8 @@ const getRowActionAttributes = (row) => {
   if (row && row.action) {
     if (typeof(row.action.onAction) === 'function') {
       return { onClick: () => row.action.onAction(row) };
-    } else {
-      return { href: row.action.href };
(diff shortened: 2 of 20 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "post-compile": "rimraf dist/*.test.* dist/**/*.test.* dist/**/*.stories.* dist/docs dist/json dist/assets"
   },
   "dependencies": {
-    "@ukhomeoffice/cop-react-components": "^0.8.1",
+    "@ukhomeoffice/cop-react-components": "^0.8.2",
```
keenan218, line 19: The current version of cop-react-components is 1.1.2, should this be reflected here?
```suggestion
    "@ukhomeoffice/cop-react-components": "^1.1.2",
```

src/components/SummaryList/helpers/getRowActionAttributes.js
```diff
@@ -2,8 +2,8 @@ const getRowActionAttributes = (row) => {
   if (row && row.action) {
     if (typeof(row.action.onAction) === 'function') {
       return { onClick: () => row.action.onAction(row) };
-    } else {
-      return { href: row.action.href };
+    } else if (row.action.page) {
```
keenan218, line 5: Why is the `else if` needed now?

package.json
paulfgraham: Not in this one or the unit tests will fail. It's in this PR:
https://github.com/UKHomeOffice/cop-react-form-renderer/pull/59/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R19

keenan218: Ah yes, just had a look at the other PR and seen its updated.

src/components/SummaryList/helpers/getRowActionAttributes.js
keenan218: Thanks for clearing that up

PR #56, opened 2022-03-07
EAB-132 Allow strings in options data

gardnersam:


paulfgraham (commented):
The changes to the code all look good, but is there anything in the documentation around options that might need updating too? I'm thinking of the examples or component stories inside `src/docs`. 🤔

PR #64, opened 2022-03-18
EAB-142 Add custom required error messages

gardnersam:
Designs for new EAB screens require a variety of error messages to be displayed when required validation is failed. This PR adds the capability to specify custom messages for this purpose in the JSON.

### DIFF
--- a/src/docs/Component.stories.mdx
+++ b/src/docs/Component.stories.mdx
@@ -48,10 +48,17 @@ Whether or not the field is required (defaults to `false`).
 If this is `false`, the label will have an '(optional)' suffix; if this is `true`,
 an error will be shown if the field is not populated.
 
-### `customError : string`
+### `custom_error : object`
 *Optional, and only applicable to <Link href="/?path=/docs/f-editable-components">editable components</Link>*
 
-The error message that should be displayed if the `required` validation is failed.
+The error message that should be displayed if a validation check is failed, currently usable on the `required` validator.
+
+Structured as: `"custom_errors": [
+  {
+    "type": "required",
+    "message": "Custom message here"
+  }
+]`
 
 ### `readonly: boolean`
 *Optional, and only applicable to <Link href="/?path=/docs/f-editable-components">editable components</Link>*
--- a/src/docs/Component.stories.mdx
+++ b/src/docs/Component.stories.mdx
@@ -48,6 +48,11 @@ Whether or not the field is required (defaults to `false`).
 If this is `false`, the label will have an '(optional)' suffix; if this is `true`,
 an error will be shown if the field is not populated.
 
+### `customError : string`
+*Optional, and only applicable to <Link href="/?path=/docs/f-editable-components">editable components</Link>*
+
+The error message that should be displayed if the `required` validation is failed.
+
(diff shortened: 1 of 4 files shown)


src/docs/Component.stories.mdx
```diff
@@ -48,6 +48,11 @@ Whether or not the field is required (defaults to `false`).
 If this is `false`, the label will have an '(optional)' suffix; if this is `true`,
 an error will be shown if the field is not populated.
 
+### `customError : string`
```
paulfgraham, line 51: I'm not so keen on this being called `customError` if it's _only_ applicable to the `required` field. Maybe `customRequiredError` or perhaps an array of objects such as:
```
"customErrors": [
  { "error": "required", "message": "Use it or lose it" }
]
```

paulfgraham (changes requested):

src/utils/Validate/validateRequired.js
```diff
@@ -16,6 +16,14 @@ const validateRequired = (value, label = '') => {
     }
   }
   if (!hasValue) {
+    if(customErrors && Array.isArray(customErrors)){
```
paulfgraham, line 19: You don't really need the first part of this as `Array.isArray(undefined|null)` will simply evaluate to `false`.
```suggestion
    if (Array.isArray(customErrors)) {
```

src/utils/Validate/validateRequired.js
```diff
@@ -16,6 +16,14 @@ const validateRequired = (value, label = '') => {
...
+      const result = customErrors.filter(error => {
+        return error.type === 'required';
+      })
+      if(result && result.length > 0 && result[0].message){
```
paulfgraham, line 23: Purely aesthetically...
```suggestion
      if (result && result.length > 0 && result[0].message) {
```

src/utils/Validate/validateRequired.js
```diff
@@ -16,6 +16,14 @@ const validateRequired = (value, label = '') => {
     }
   }
   if (!hasValue) {
+    if(customErrors && Array.isArray(customErrors)){
+      const result = customErrors.filter(error => {
+        return error.type === 'required';
+      })
```
paulfgraham, line 22:
```suggestion
      });
```

src/docs/Component.stories.mdx
gardnersam: I've swapped to the array of objects as I think  that will be more extensible in the future

paulfgraham (commented):
Just a couple of formatting things. Otherwise, it looks good...

src/utils/Validate/validateRequired.test.js
```diff
@@ -45,6 +45,12 @@ describe('utils', () => {
       it('should use a default label when none is specified', () => {
         expect(validateRequired(undefined, undefined)).toEqual('Field is required');
       });
+      it('should use a custom label when one is provided', () => {
```
paulfgraham, line 48: Sorry, missed this... you mean a custom _error_, right? Same for the one below...

paulfgraham (approved):
Thanks for making the requested changes, @gardnersam, even with me being picky. 😁

PR #67, opened 2022-04-04
Feature/cop 10344/caseview v3

keenan218:
### Description
This change has been done to support changes in cop to display read-only versions of submitted forms when in CaseView. The styling has been applied to match what was outlined in the figma documentation containing the initial designs of CaseView.

https://support.cop.homeoffice.gov.uk/browse/COP-10344

### To test
Testing that the functionality works can be achieved by viewing the changes in storybook (`yarn storybook`) and looking in each of the respected sections for Check Your Answers, Form Renderer and Summary List. At the bottom there will be a read-only view which contains no change actions, or no lines.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "2.0.1",
+  "version": "2.1.0",
   "private": false,
   "scripts": {
     "clean": "rimraf dist",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "private": false,
   "scripts": {
     "clean": "rimraf dist",
--- a/src/components/CheckYourAnswers/CheckYourAnswers.jsx
+++ b/src/components/CheckYourAnswers/CheckYourAnswers.jsx
@@ -1,5 +1,9 @@
 // Global imports
-import { ErrorSummary, LargeHeading, MediumHeading } from '@ukhomeoffice/cop-react-components';
+import {
+  ErrorSummary,
+  LargeHeading,
+  MediumHeading
+} from '@ukhomeoffice/cop-react-components';
 import PropTypes from 'prop-types';
 import React, { Fragment, useEffect, useState } from 'react';
 
@@ -22,7 +26,11 @@ const CheckYourAnswers = ({
   onAction,
   onRowAction,
   hide_page_titles,
-  hide_actions
+  hide_actions,
+  hide_title,
+  summaryListClassModifiers,
+  noAction,
+  invertWeight
 }) => {
   const [pages, setPages] = useState([]);
   const [errors, setErrors] = useState([]);
@@ -32,19 +40,29 @@ const CheckYourAnswers = ({
       const rows = Utils.CheckYourAnswers.getRows(page, onRowAction);
       return rows.map((row, index) => ({
         ...row,
-        value: <Answer key={`${pageIndex}_${index}`} value={row.value} component={row.component} />
+        value: (
+          <Answer
+            key={`${pageIndex}_${index}`}
+            value={row.value}
+            component={row.component}
+          />
+        )
       }));
     };
 
-    setPages(_pages.map((page, index) => {
-      const rows = getRows(page, index);
-      return rows.length > 0 ? { ...page, rows } : null;
-    }).filter(p => !!p));
+    setPages(
+      _pages
+        .map((page, index) => {
+          const rows = getRows(page, index);
+          return rows.length > 0 ? { ...page, rows } : null;
+        })
+        .filter((p) => !!p)
+    );
   }, [_pages, onRowAction, setPages]);
 
   const listMarginBottom = hide_page_titles ? 0 : DEFAULT_MARGIN_BOTTOM;
   const isLastPage = (index) => {
-    return index === pages.length -1;
+    return index === pages.length - 1;
   };
 
   const onError = (errors) => {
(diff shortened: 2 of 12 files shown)


src/components/CheckYourAnswers/CheckYourAnswers.test.js
```diff
@@ -129,6 +129,17 @@ describe('components', () => {
       checkRow(status, 'Are you a civil servant?', 'Yes', true);
     });
 
+    it('should show no title if is set to hidden', async () => {
+      await act(async () => {
+        render(<CheckYourAnswers pages={PAGES} onRowAction={ON_ROW_ACTION} onAction={ON_ACTION} hide_title={true}/>, container);
+      });
+      const cya = checkCYA(container);
+      const title = cya.childNodes[0];
```
gardnersam, line 137: I don't think this is / should be the title, it looks like it should be the output from the pages.map() starting on line 78.

src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -22,7 +26,11 @@ const CheckYourAnswers = ({
   onAction,
   onRowAction,
   hide_page_titles,
-  hide_actions
+  hide_actions,
+  hide_title,
+  summaryListClassModifiers,
+  noAction,
```
gardnersam, line 32: This and invertWeight are purely here to be passed on to the summary list so wonder if that could be in the name like 'summaryListClassModifiers'. Not a deal breaker but I think it's confusing having hide_actions and noAction

src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -177,15 +185,23 @@ FormRenderer.propTypes = {
   }),
   classBlock: PropTypes.string,
   classModifiers: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
-  className: PropTypes.string
+  className: PropTypes.string,
+  hide_title: PropTypes.bool,
+  summaryListClassModifiers: PropTypes.string,
```
gardnersam, line 190: same comment as above, should have an array allowed here

src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -177,15 +185,23 @@ FormRenderer.propTypes = {
...
 
 FormRenderer.defaultProps = {
   type: FormTypes.HUB,
   components: [],
   pages: [],
   classBlock: DEFAULT_CLASS,
-  classModifiers: []
+  classModifiers: [],
+  hide_title: false,
+  summaryListClassModifiers: null,
```
gardnersam, line 202: would be more consistent to default this to [] like classModifiers

src/components/SummaryList/SummaryList.jsx
```diff
@@ -12,47 +12,72 @@ import './SummaryList.scss';
...
-          <dt className={classes('key')}>{row.key}</dt>
-          <dd className={classes('value')}>{row.value}</dd>
-          <dd className={classes('actions')}>
-            {row.action && <RowAction row={row} />}
-          </dd>
-        </div>
-      ))}
+    <dl {...attrs} className={classes()}>
+      {rows.map((row) =>
+        invertWeight ? (
```
gardnersam, line 26: I don't really understand what invertWeight is supposed to do, but the only difference in the code seems to be adding the two styles `caseview__key` and `caseview__value`. So could the code duplication be reduced by determining the classes upfront ie `const keyStyle = invertWeight ? 'key caseview__key' : 'key'` then have a single block of code which takes `keyStyle` and a simular `valueStyle` for the input to classes like `<dt className={classes(keyStyle)}>{row.key}</dt>`

src/components/SummaryList/SummaryList.scss
```diff
@@ -1,9 +1,17 @@
...
+@import 'node_modules/govuk-frontend/govuk/components/summary-list/_summary-list';
 
 .govuk-summary-list__actions {
   .govuk-link {
     color: govuk-colour('blue');
     cursor: pointer;
   }
 }
+
+.caseview__key {
```
gardnersam, line 11: now i've got down here it seems like the `invertWeight` prop on SummaryList is to adjust the font weight so maybe could be named clearer to reflect that

src/components/SummaryList/SummaryList.test.js
```diff
@@ -20,14 +18,27 @@ describe('components', () => {
...
       expect(key.classList).toContain(`${DEFAULT_CLASS}__key`);
       expect(value.tagName).toEqual('DD');
       expect(value.classList).toContain(`${DEFAULT_CLASS}__value`);
       expect(actions.tagName).toEqual('DD');
       expect(actions.classList).toContain(`${DEFAULT_CLASS}__actions`);
-      return [ key, value, actions ];
+      return [key, value, actions];
+    };
+
+    const checkRowNoActions = (summaryList, index) => {
```
gardnersam, line 31: this function doesn't check that there are no actions, if anything it seems to assume there are actions and return them

src/components/SummaryList/SummaryList.test.js
```diff
@@ -109,6 +142,77 @@ describe('components', () => {
...
+      );
+      const summaryList = checkSummaryList(container, ID);
+      expect(summaryList.childNodes.length).toEqual(ROWS.length);
+      ROWS.forEach((row, index) => {
+        const [key, value] = checkRowNoActions(summaryList, index);
+        expect(key.textContent).toEqual(row.key);
+        expect(value.childNodes.length).toEqual(1);
+        const valueDiv = value.childNodes[0];
+        expect(valueDiv.tagName).toEqual('DIV');
+        expect(valueDiv.textContent).toEqual(VALUES[index]);
```
gardnersam, line 173: There doesn't seem to be a check that actions are hidden, could we check that the length of the childNodes array for the row is 2? (if that's the right answer)

gardnersam (changes requested):
Looks good overall, few thoughts on some of the implementation details

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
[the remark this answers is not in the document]
keenan218: Understood, will change thanks

src/components/CheckYourAnswers/CheckYourAnswers.jsx
keenan218: Point taken, could be confusing having both pretty similar, will look into this

src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -77,13 +113,21 @@ CheckYourAnswers.propTypes = {
   onAction: PropTypes.func.isRequired,
   onRowAction: PropTypes.func.isRequired,
   hide_page_titles: PropTypes.bool,
-  hide_actions: PropTypes.bool
+  hide_actions: PropTypes.bool,
+  hide_title: PropTypes.bool,
+  summaryListClassModifiers: PropTypes.string,
```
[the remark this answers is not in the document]
keenan218: Will change this to reflect how the PropTypes are in SummaryList

src/components/FormRenderer/FormRenderer.jsx
keenan218: Yep i agree, will get that changed

src/components/SummaryList/SummaryList.scss
keenan218: invertWeight has been changed to invertFontWeight, used when swapping a bold font to normal font, hope this is better.

keenan218: After re-checking the figma it has changed once more and now it seems the invertWeight functionality is now not needed..

src/components/CheckYourAnswers/CheckYourAnswers.test.js
keenan218: A more suitable name has been given to this rather than title

src/components/CheckYourAnswers/CheckYourAnswers.jsx
keenan218: I have given noAction a more suitable name of noChangeAction so it hopefully doesn't clash with hide_actions too much

src/components/SummaryList/SummaryList.test.js
keenan218: We have now included a check for the length of the childNodes array being only 2

src/components/SummaryList/SummaryList.test.js
keenan218: Have altered the function so it checks that there are no actions and does not return any actions.

gardnersam (approved):
Looks good!

PR #72, opened 2022-04-19
COP-10564: Add hook for overriding component

paulfgraham:
### Descripton
Added a hook for overriding components. The override function is passed the `wrap` parameter where a value of `false` means it _shouldn't_ be wrapped and a value of `true` means the function should decide whether or not it's should be within a `FormGroup` and wrap it if necessary.

https://support.cop.homeoffice.gov.uk/browse/COP-10564

### To test


gardnersam (approved):
Looks good to me

PR #73, opened 2022-04-20
Feature/cop 7991 add tasklist navigation to sections

gardnersam:
Add the functionality to move from the tasklist hub to individual sections, through each page in the section and finally end on a CYA page for the section. Adds a new action type to allow returning from a CYA screen to the tasklist rather than submitting the form. 

The inter-section navigation is reliant on the JSON for each page specifying which page should come next. This seemed like the simplest method and allows versatility in the form of optionally adding CYA as the final page. However this does lead to some duplication as each section specifies the pages it includes as well which is required to populate the CYA pages with the correct fields.

### DIFF
--- a/src/components/TaskList/Task.jsx
+++ b/src/components/TaskList/Task.jsx
@@ -23,7 +23,7 @@ const Task = ({ task, onClick }) => {
   return (
     <li className={classes('item')}>
       <span className={classes('task-name')}>
-        {linkActive ? <Link onClick={() => onClick(task.firstPage)}>{task.name}</Link> : task.name}
+        {linkActive ? <Link onClick={() => onClick(task)}>{task.name}</Link> : task.name}
       </span>
       <TaskState state={currentState} />
     </li>
@@ -33,7 +33,7 @@ const Task = ({ task, onClick }) => {
 Task.propTypes = {
   task: PropTypes.shape({
     name: PropTypes.string.isRequired,
-    firstPage: PropTypes.string.isRequired,
+    pages: PropTypes.array.isRequired,
     state: PropTypes.string.isRequired,
   }).isRequired,
   onClick: PropTypes.func.isRequired,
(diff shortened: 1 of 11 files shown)


src/components/TaskList/Task.jsx
```diff
@@ -33,7 +33,7 @@ const Task = ({ task, onClick }) => {
 Task.propTypes = {
   task: PropTypes.shape({
     name: PropTypes.string.isRequired,
-    firstPage: PropTypes.string.isRequired,
+    pages: PropTypes.array.isRequired,
```
keenan218, line 36: Moreso just for my understanding, why has `firstPage` now become just `pages`

src/json/taskList.json
```diff
@@ -0,0 +1,228 @@
...
+          {
+            "value": "air",
+            "label": "Air"
+          }
+        ]
+      },
+      "required": true
+    }
+  ],
+  "pages": [
```
keenan218, line 62: Ah, is this what the `pages` everywhere else is referencing

src/components/TaskList/Task.jsx
gardnersam: At first I thought all we needed to know was the start point for a task. So we could just move to that and let the FormRenderer logic take care of the rest. Problem was with the CYA pages at the end of a task, existing code just displays all the pages that exist within the form as a whole. So needed to manage that better so that you only see the answers for the pages that are actually in form. 

Question was then how do you know what pages are in the task if you are just specifying the order within the pages them selves. So the conclusion I arrived at was we have to document them with the definition of the task itself. So now when a CYA page is rendered it is passed just the pages that are given in the pages prop

keenan218: Ah okay makes sense, thanks for the explanation :)

keenan218 (approved):
Looks good

PR #74, opened 2022-04-27
COP-9885 Add task state management

gardnersam:
### Description
Add management of the state of each task. The task state can be 'Cannot Start', 'Not Started, 'In Progress' or 'Complete'. This information is stored in the backend and updated as the user progresses through the task list. Details for when tasks should be in each state and progress to the next one are covered in the jira ticket. 

The updates to `FormRenderer.jsx` largely cover the updating of states when the user either moves to the next page or completes a section. This uses the new helper function `getUpdatedSectionStates()` to determine each tasks state and an updated `getSubmissionStatus()` to add state information to the submitted data object.

### Testing
Covered by unit tests


keenan218 (approved):
Looks Good

PR #75, opened 2022-04-27
COP-10558: Modularise conditional show logic

paulfgraham:
### Description
Moved the `meetsCondition` utility function into a `Condition` directory and added in a new utility method for checking that all conditions have been met, which had already been done in two other places (and would have been being done in a third without this refactor).

### To test
This is covered by the accompanying unit tests, which have 100% coverage.


paulfgraham:
The `meetsCondition.test.js` file has simply been moved from its previous location but then also wrapped within a `describe('Container', () => { });` block... which is why it looks like a whole brand new file.

https://github.com/UKHomeOffice/cop-react-form-renderer/pull/75/files#diff-0249b131ff6185aa03c70d6aff83382437931d0250af970607e0837e34873460

charles474 (approved):
Thanks for unwrapping the wrapped `meetsCondition.test.js` test

PR #82, opened 2022-05-05
COP-10558: Add Container component

paulfgraham:
### Description
Added the `Container` component, although it's within the `FormComponent.jsx` file to avoid circular dependencies - something that was causing a problem when the Container was in its own file.

### To test


src/components/FormComponent/FormComponent.jsx
```diff
@@ -3,13 +3,78 @@ import PropTypes from 'prop-types';
...
 
 // Local imports
-import useHooks from '../../hooks/useHooks';
-import useRefData, { STATUS_COMPLETE } from '../../hooks/useRefData';
+import { STATUS_COMPLETE, useHooks, useRefData } from '../../hooks';
+import { ComponentTypes } from '../../models';
 import Utils from '../../utils';
 
+export const DEFAULT_CONTAINER_CLASS = 'hods-form-container';
+const Container = ({
```
paulfgraham, line 11: As mentioned in the description for this PR, this _would_ be in its own file, except that putting it there resulted in a circular dependency and a failure in the unit tests. The circular dependency didn't seem to be a problem while running and it behaved itself just fine... but the unit tests complained as soon as this component tried to render a `<FormComponent>` element of its own.

`FormComponent => Container => FormComponent`

PR #83, opened 2022-05-05
COP-10558: Support containers on Check your answers screen

paulfgraham:
### Description
Support `Container` components on the **Check your answers** screen. Presently, each editable component within a `Container` will be handled exactly the same as if it was at the top level of the form.

### To test


charles474 (approved):
LGTM

PR #84, opened 2022-05-05
COP-10558: Support containers on form pages

paulfgraham:
### Description
When a form page includes a `Container` component, there is a need to set up the nesting that comes with that. These changes handle that setup.

In addition, removed the utility function for getting editable components on a Page (and its accompanying unit tests), which has been superseded by the utility function for getting editable components within a Container (which also supports nested Containers).

### To test


charles474 (approved):
LGTM

PR #85, opened 2022-05-06
COP-10567: Fix cerberus import issue

paulfgraham:
### Description
Fixed an issue with the `FormComponent.jsx` file, which wasn't importing `React` despite including JSX markup. This meant that importing version 2.6.0 and trying to run anything that included the `FormComponent` would throw an exception - i.e., placing a `<FormRenderer>` element on a page or within another component would fail, which is rather fatal.


charles474 (approved):
LGTM

Nice one getting to the root of the problem 😏

PR #87, opened 2022-05-13
COP-10567: Add support for "file" component type

paulfgraham:
### Description
Added support for the component type "file", which will render a `<FileUpload />` element from the component library.

Also added unit tests and to the Storybook documentation for this new type.

https://support.cop.homeoffice.gov.uk/browse/COP-10567

### To test
Covered by unit tests.


gardnersam (approved):
Looks good!

PR #93, opened 2022-05-18
Feature/c op 2120 cya validation

georgepemberton1998:
In the current build validation occurs on required fields during the "check your answers" page even when the page that they belong to has not been displayed due to conditional logic on the page itself. A work around to this is possible by passing the same conditional "show-when" on the components as the page has however this is not ideal and quite difficult to document.

The below changes adds an additional step when validating a page to first check whether or not the page was actually displayed to the user.

Jira ticket link - https://support.cop.homeoffice.gov.uk/browse/CO-2120

### DIFF
--- a/src/utils/Validate/validatePage.js
+++ b/src/utils/Validate/validatePage.js
@@ -10,12 +10,13 @@ import showFormPage from '../FormPage/showFormPage';
  * @returns An array containing all of the errors.
  */
  const validatePage = (page) => {
-  if (Array.isArray(page.components) && showFormPage(page, page.formData)) {
+  if (Array.isArray(page?.components) && showFormPage(page, page.formData)) {
     return page.components.reduce((errors, component) => {
       return errors.concat(validateComponent(component, page.formData));
     }, []).filter(e => !!e).flat();
   }
   return [];
 };
+
 
 export default validatePage;
--- a/src/utils/Validate/validatePage.js
+++ b/src/utils/Validate/validatePage.js
@@ -1,5 +1,6 @@
 // Local imports
 import validateComponent from './validateComponent';
+import showFormPage from '../FormPage/showFormPage';
 
 /**
  * Validate all of the components on a page.
@@ -7,8 +8,8 @@ import validateComponent from './validateComponent';
  * @param {object} formData The data to use that holds this components' values.
  * @returns An array containing all of the errors.
  */
-const validatePage = (components, formData) => {
-  if (Array.isArray(components)) {
+const validatePage = (components, formData, page) => {
+  if (Array.isArray(components) && showFormPage(page, formData)) {
     return components.reduce((errors, component) => {
       return errors.concat(validateComponent(component, formData));
     }, []).filter(e => !!e).flat();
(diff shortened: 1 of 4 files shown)


src/utils/Validate/validatePage.js
```diff
@@ -1,19 +1,22 @@
...
  * @param {Array} components The components to validate.
  * @param {object} formData The data to use that holds this components' values.
  * @returns An array containing all of the errors.
  */
-const validatePage = (components, formData) => {
-  if (Array.isArray(components)) {
-    return components.reduce((errors, component) => {
-      return errors.concat(validateComponent(component, formData));
+ const validatePage = (page) => {
+  if (Array.isArray(page?.components) && showFormPage(page, page.formData)) {
```
paulfgraham, line 13: If you flip these conditions around, you won't need the optional chaining as a `null` or `undefined` page won't be shown:
```suggestion
  if (showFormPage(page, page.formData) && Array.isArray(page.components)) {
```

src/utils/Validate/validatePage.js
```diff
@@ -1,19 +1,22 @@
...
+    return page.components.reduce((errors, component) => {
+      return errors.concat(validateComponent(component, page.formData));
     }, []).filter(e => !!e).flat();
   }
   return [];
 };
 
+
```
paulfgraham, line 21: Not sure why this extra blank line has been introduced...

paulfgraham (changes requested):
A few things I think could do with adjusting.

src/utils/Validate/validatePage.js
georgepemberton1998: I made this change however the unit test now fails due to formData being null..

`  ● utils › Validate › page › should return no error when the components array is null

    TypeError: Cannot read properties of null (reading 'formData')

      11 |  */
      12 |  const validatePage = (page) => {
    > 13 |   if (showFormPage(page, page.formData) && Array.isArray(page.components)) {
         |                               ^
      14 |     return page.components.reduce((errors, component) => {
      15 |       return errors.concat(validateComponent(component, page.formData));
      16 |     }, []).filter(e => !!e).flat();

      at validatePage (src/utils/Validate/validatePage.js:13:31)
      at Object.<anonymous> (src/utils/Validate/validatePage.test.js:16:16)
`

I can avoid this by updated the first unit test to the following 
`      it('should return no error when the components array is null', () => {
        const PAGE = {
          components: null,
          formData: {
            alpha: 'alpha.smith@digital.homeoffice.gov.uk',
            bravo: 'bravo.jones@digital.homeoffice.gov.uk'
          }
        };
        expect(validatePage(PAGE).length).toEqual(0);
      });`
      
 Let me know if you would prefer that

georgepemberton1998: I have pushed a new commit addresses this and other comments - let me know if you would prefer to keep the optional chaining over editing the unit test and I can revert that part.

paulfgraham (approved):
Looks good, @georgepemberton1998. Thank you for making the requested changes. 👍

PR #95, opened 2022-05-20
Feature/cop 10715/group row changes

aliktb:
**Description**
Added option for a group of rows in CYA screen to navigate to the question page using a single change link. This can be specified in the form json in the `cya` object under a `groups` property

https://support.cop.homeoffice.gov.uk/browse/COP-10715

**To test**


src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -1,132 +1,191 @@
...
+    const groupsList = () => {
+
+      let groupId = [];
+      
+      groups !== undefined &&
+        groups.length && groups.forEach((group) => {
+          const groupedPageId = group.pageId;
+          groupId.push(groupedPageId);
+        })
+      console.log('object');
```
gardnersam, line 84: don't think this should still be here

src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -1,132 +1,191 @@
...
+  
+        setListOfGroups(groupId);
+    }
+
+    groupsList();
+
```
gardnersam, line 90: this function feels overly spaced out compared to others in the file

src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -1,132 +1,191 @@
...
+  const onError = (errors) => {
+    setErrors(errors);
+  };
+
+  useEffect(() => {
+
+    const groupsList = () => {
+
+      let groupId = [];
```
gardnersam, line 77: think this would be better named as `groupIds` as its a list of multiple things

src/components/CheckYourAnswers/CheckYourAnswers.scss
```diff
@@ -1,2 +1,12 @@
...
+
+.govuk-link {
+    @include govuk-font($size: 19);
+    color: govuk-colour('blue');
+    cursor: pointer;
+  }
+
+.group-title{
+    padding-right: 75px;
+}
```
gardnersam, line 12: missing newline

src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -1,132 +1,191 @@
...
+  const groupedFields = (pageId) => {
+  
+    let groupAnswer;
+    groups !== undefined &&
+      groups.length && pageId !== undefined && groups.forEach((group) => {
+        if (group.pageId === pageId) {
+          groupAnswer = group;
+        }
+      })
+    console.log(groupAnswer);
```
gardnersam, line 108: rogue console.log

src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -1,132 +1,191 @@
...
+    return (groupAnswer);
```
gardnersam, line 109: are the brackets needed here

src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -1,132 +1,191 @@
...
+    
+  }, [setListOfGroups, groups]);
+
+  const isGroup = (pageId) => {
+
+    return (listOfGroups.includes(pageId));
+    
+  }
+
+  const groupedFields = (pageId) => {
```
gardnersam, line 99: I think this function takes a pageId and finds a group that also has that pageId, wonder if it could be better names to reflect that? It doesn't seem to involve fields at any point. Maybe something like getGroupForPage?

src/components/SummaryList/GroupAction.jsx
```diff
@@ -0,0 +1,42 @@
...
+import React, { useEffect, useState } from 'react';
+import { Link } from '@ukhomeoffice/cop-react-components';
+import PropTypes from 'prop-types';
+
+// Local imports
+import getGroupActionAttributes from './helpers/getGroupActionAttributes';
+
+const GroupAction = ({ group }) => {
+  const [attrs, setAttrs] = useState({});
+  console.log(group);
```
gardnersam, line 11: think this can be removed

src/components/CheckYourAnswers/CheckYourAnswers.test.js
```diff
@@ -134,7 +137,8 @@ describe('components', () => {
         render(<CheckYourAnswers pages={PAGES} onRowAction={ON_ROW_ACTION} onAction={ON_ACTION} hide_title={true}/>, container);
       });
       const cya = checkCYA(container);
-      const [ names ] = cya.childNodes;
+      const [cyaChildNode] = cya.childNodes;
+      const names = cyaChildNode.childNodes[0];
       expect(names.tagName).toEqual('DL');
       expect(names.classList).toContain(`govuk-!-margin-bottom-${DEFAULT_MARGIN_BOTTOM}`);
       const [ firstName, surname ] = names.childNodes;
```
gardnersam, line 144: Could use some tests checking that the grouped rows are getting rendered properly when they should be

src/json/groupOfRow.json
```diff
@@ -0,0 +1,376 @@
...
+  ],
+  "hub": {
+    "format": "CYA"
+  },
+  "cya": {
+    "hide_page_titles": true,
+    "hide_actions": true,
+    "groups": [
+      {
+        "pageId": "dateLocDet",
```
gardnersam, line 342: I don't understand how this relates to the page. Should this show the fields from the page or the fields in the fields array?

src/json/groupOfRow.json
```diff
@@ -0,0 +1,376 @@
...
+        "title" :"Date and location",
+        "type": "fields",
+        "fields": [
```
gardnersam, line 345: I don't think the code checks this array anywhere. I think it handles using the pageId but not these fields. We should be able to either group by page or field.

gardnersam (changes requested):

src/components/CheckYourAnswers/CheckYourAnswers.jsx
aliktb: yep, well spotted

src/components/CheckYourAnswers/CheckYourAnswers.jsx
aliktb: agreed, will remove white space

src/components/CheckYourAnswers/CheckYourAnswers.jsx
aliktb: Good suggestion

src/components/CheckYourAnswers/CheckYourAnswers.jsx
aliktb: nope, they aren't

src/components/CheckYourAnswers/CheckYourAnswers.jsx
aliktb: yep good idea. I always have a hard time trying to name variables and functions

src/components/SummaryList/GroupAction.jsx
aliktb: agreed

src/components/CheckYourAnswers/CheckYourAnswers.scss
aliktb: well spotted

src/components/CheckYourAnswers/CheckYourAnswers.jsx
aliktb: yep must've slipped in accidentally

src/json/groupOfRow.json
aliktb: Yes, you're right. I think that was the initial idea from the json but the requirements are that a group will only ever belong on one page so we don't need to check the fields. Rather, we just need to specify the page.

src/components/CheckYourAnswers/CheckYourAnswers.test.js
aliktb: I've added a test to check for a case with a group of questions

src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -1,132 +1,186 @@
...
+                <MediumHeading>{page.title}</MediumHeading>
+              )}
+              {(isGroup(page.id)) && (currentGroup.title ? (
+                <div className='group-title'>
+                    <MediumHeading>{currentGroup.title}</MediumHeading>
+              </div>
+              ) : (
+                <div className='group-title'>
+                  <MediumHeading>{page.title}</MediumHeading>
+                </div>
```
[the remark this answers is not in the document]
gardnersam: could even extract a title variable maybe to avoid repeating `<MediumHeading>` ?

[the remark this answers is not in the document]
aliktb: good call, I think I've done something to that effect now

gardnersam (approved):
Looks great!

PR #107, opened 2022-06-06
swapped gds transport for roboto in tasklist

aliktb:
**Description**
Removed GDS Transport font from tasklist


keenan218 (approved):
Approved

PR #110, opened 2022-06-07
COP-10873: Add collection validation

paulfgraham:
### Description
Added validation for collections and also refactored the `validateComponent` and `validateContainer` methods so they're in their own files, rather than all bundled into one.

### To test

### DIFF
--- a/src/utils/Validate/validateCollection.js
+++ b/src/utils/Validate/validateCollection.js
@@ -1,0 +1,24 @@
+// Local imports
+import validateContainer from './validateContainer';
+
+const validateCollection = (collection, items) => {
+  const errors = [];
+  if (collection && Array.isArray(collection.item) && Array.isArray(items)) {
+    items.forEach((item, index) => {
+      const full_path = `${collection.full_path || collection.fieldId}[${index}]`;
+      const container = {
+        full_path,
+        components: collection.item.map(component => {
+          return {
+            ...component,
+            full_path: `${full_path}.${component.fieldId}`
+          };
+        })
+      };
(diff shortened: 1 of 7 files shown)


src/utils/Validate/validateComponent.test.js
```diff
@@ -2,11 +2,7 @@
...
 import validateComponent from './validateComponent';
 
-describe('utils', () => {
-
-  describe('Validate', () => {
-
-    describe('component', () => {
+describe('utils.Validate.Component', () => {
 
       const setup = (id, type, label, required, additionalValidation) => {
```
paulfgraham, line 7: Note that I haven't sorted out the indenting on most of the lines in here... because it would mean ALL of the lines are changed.

src/utils/Validate/validateCollection.js
```diff
@@ -0,0 +1,24 @@
+// Local imports
+import validateContainer from './validateContainer';
+
+const validateCollection = (collection, items) => {
+  const errors = [];
+  if (collection && Array.isArray(collection.item) && Array.isArray(items)) {
+    items.forEach((item, index) => {
+      const full_path = `${collection.full_path || collection.fieldId}[${index}]`;
```
gardnersam, line 8: are we guaranteed that the order of the fields in the collection will match the order of their corresponding data in items?

gardnersam: Thanks for the explanation, resolving

gardnersam (approved):
Looks good, thanks for the explanations

PR #115, opened 2022-06-08
COP-10907 - reworked optional tag declaration

keenan218:
Changed how '(optional) is displayed on forms. When one field is required in a row where multiple fields are displayed the row will not show optional. optional will only be displayed when all fields in a row are optional. 

The json for forms will have to declare the state of required.


aliktb (approved):
Looks good

PR #116, opened 2022-06-08
Add Collection component

paulfgraham:
### Description
Added a `Collection` component and, in the midst of this, moved `Container` to its own file, rather than having all 3 inside of `FormComponent`.

This refactor also leverages the Meta utils now, most of the code from which had previously sat inside of `FormComponent.jsx`.

### To test
Covered by accompanying unit tests (not yet present for the `Collection` component itself, but will be added soon.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "2.9.1",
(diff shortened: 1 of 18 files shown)


charles474 (approved):
Approved when the changes were showing 5 files.

PR #120, opened 2022-06-09
COP-10873: Improve collection validation

paulfgraham:
### Description
Added support for validation of conditionally-shown components within collections. Previously, any conditionally-shown components within collections were simply considered _not_ to be shown, which meant they weren't validated.

### To test
Within the Form Renderer Sandbox, set up a form with a collection that contains a conditionally-shown field. Make sure the validation on that field kicks in when it is shown and the validation is ignored when it's _not_ shown.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "3.0.1",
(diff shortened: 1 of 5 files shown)


Shaks (approved):
lgtm

PR #123, opened 2022-06-13
COP-10947: Refactor useHooks

paulfgraham:
### Description
`useHooks` is now a pass-through for the `HooksContext`, which makes all of the hooks instance-specific.

Updated a bunch of unit tests so that they're now wrapping testable components inside of a `<HooksContextProvider />` element where they have `useHooks` internally.

https://support.cop.homeoffice.gov.uk/browse/COP-10947

### To test
Covered by accompanying and updated unit tests.

### DIFF
--- a/src/hooks/useHooks.js
+++ b/src/hooks/useHooks.js
@@ -1,50 +1,11 @@
-export const ALLOWED_HOOKS = [
-  'onFormComplete',
-  'onFormLoad',
-  'onGetComponent',
-  'onPageChange',
-  'onRequest',
-  'onSubmit'
-];
+// Global imports
+import { useContext } from 'react';
 
-const DEFAULT_HOOKS = {
-  onFormComplete: () => {},
-  onFormLoad: () => {},
-  onGetComponent: (config, wrap) => null,
-  onPageChange: (pageId) => pageId,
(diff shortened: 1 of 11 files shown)


src/hooks/useHooks.js
```diff
@@ -1,50 +1,11 @@
...
-
-export const resetHooks = () => {
-  Object.keys(hooks).forEach(removeHook);
-};
+// Local imports
+import { HooksContext } from '../context/HooksContext';
 
 const useHooks = () => {
-  return { hooks, addHook, removeHook, resetHooks };
+  return useContext(HooksContext);
```
gardnersam, line 8: why doesn't HooksContext just return this and save the need for this file?

gardnersam (approved):
Looks good

src/hooks/useHooks.js
paulfgraham: Technically speaking, you could implement line 8 in each component where you'd want to get hold of the context:
```
const { hooks } = useContext(HooksContext);
```

However, I find that a custom hook that wraps that up is just a cleaner way of doing it and it _seems_ to be the most common mechanism used for this sort of thing.

PR #124, opened 2022-06-14
Import React in ValidationContext.jsx

fisherrj2:
While running tests for cop-ui there was an error reporting that React was not defined and a stack trace ending in the ValidationContext jsx file in the renderer.  The fix was to import React from react within that file.


keenan218 (approved):
Looks good

PR #127, opened 2022-06-15
Implement regex validation for components.

LewisClark-HO:
Added the ability to define regex validation for a component in the form renderer JSON files. Custom errors for when the regex validation failed can also now be defined.

Doing so looks like:
```
...
"type": "text",
"pattern": "^[a-z]*$",
"custom_errors": [
   {
   "type": "pattern",
   "message": "Failed regex validation."
   }
]
...
```

### DIFF
--- a/src/utils/Validate/validateRegex.js
+++ b/src/utils/Validate/validateRegex.js
@@ -7,11 +7,12 @@
  * @param {string} label The label to use in the default error message.
  * @param {string} pattern The regex pattern to validate against.
  * @param {array} customErrors The array of custom errors for the component.
- * @returns An error if the email address is invalid.
+ * @returns An error if the value doesn't match the regex pattern.
  */
-const validateRegex = (value, label='', pattern, customErrors) => {
-  if (!value)
+const validateRegex = (value, label = '', pattern, customErrors) => {
+  if (!value) {
     return undefined;
+  }
   if (typeof value === 'string') {
     const regex = new RegExp(pattern);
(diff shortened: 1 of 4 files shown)


src/utils/Validate/validateRegex.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * Validates an components value against a given regex pattern.
+ * 
+ * An empty string for value is considered valid as not all components
+ * with pattern validation have to be required. 
+ * @param {*} value The value to validate.
+ * @param {string} label The label to use in the default error message.
+ * @param {string} pattern The regex pattern to validate against.
+ * @param {array} customErrors The array of custom errors for the component.
+ * @returns An error if the email address is invalid.
```
paulfgraham, line 10: Regex, rather than email address...

src/utils/Validate/validateRegex.js
```diff
@@ -0,0 +1,34 @@
...
+  if (typeof value === 'string') {
+    const regex = new RegExp(pattern);
+    if (regex.test(value)) {
+      return undefined;
+    }
+    if (Array.isArray(customErrors)) {
+      const result = customErrors.filter((error) => {
+        return error.type === 'pattern';
+      });
+      if (result && result.length > 0 && result[0].message) {
```
paulfgraham, line 24: You could do this like this:
```suggestion
      if (result?.[0]?.message) {
```

paulfgraham (changes requested):

PR #128, opened 2022-06-15
COP-10967: Use version 1.7.1 of component library

paulfgraham:
### Description
Use version 1.7.1 of component library, which fixes an issue with the `DateInput` component.

https://support.cop.homeoffice.gov.uk/browse/COP-10967

### To test
The following was broken in the COP React Form Renderer:
1. Set up the JSON to have a date field;
2. Set up the initial data to have a valid value for that date field;
3. Switch to the Preview tab and you'll see the value appropriately showing in the date field;
4. Go back to the initial date and change to another valid value for the date field;
5. Switch back to the Preview tab and you'll now see the value in the date field flicking between the old and new values.

This should now be fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "post-compile": "rimraf dist/*.test.* dist/**/*.test.* dist/**/*.stories.* dist/docs dist/assets"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


gardnersam (approved):
Looks good to me

PR #130, opened 2022-06-16
Add unit tests for Collection component

paulfgraham:
### Description
The `Collection` component was missing unit tests. This PR adds a bunch.

### To test
Nothing to separately test - this is adding unit tests.


gardnersam (approved):
As it's just tests not sure if the package version should be bumped or not?

paulfgraham:
Agreed with @gardnersam that the version shouldn't be bumped - it doesn't qualify as a patch release, according to https://semver.org/.

PR #136, opened 2022-06-20
Bug/fix interpolation sub

kam-cop:


src/utils/Validate/validatePage.js
```diff
@@ -1,6 +1,7 @@
 // Local imports
 import validateComponent from './validateComponent';
 import showFormPage from '../FormPage/showFormPage';
+import { Utils } from '@ukhomeoffice/cop-react-components';
```
paulfgraham, line 4: This is a global import and should appear at the top.

paulfgraham, line 4: This is a global import and should appear at the top.

src/components/FormPage/FormPage.jsx
```diff
@@ -37,6 +37,9 @@ const FormPage = ({
   };
 
   const classes = Utils.classBuilder(classBlock, classModifiers, className);
+  page.actions = page.actions?.map((action) => {
```
paulfgraham, line 40: The thing with this approach is that it's also not _truly_ dynamic. It'll happen when you first render a page but won't  re-render until you navigate back to the page again - i.e., if you want a button label to depend on something on the _current_ page, this mechanism won't work.

Having said that, I think it's highly unlikely that a page's button label would _actually_ need to be interpolated.

src/utils/Validate/validatePage.js
```diff
@@ -10,7 +11,9 @@ import showFormPage from '../FormPage/showFormPage';
 const validatePage = (page) => {
   if (showFormPage(page, page.formData) && Array.isArray(page.components)) {
     return page.components.reduce((errors, component) => {
-      return errors.concat(validateComponent(component, page.formData, page.formData));
+      return errors.concat(validateComponent(component, page.formData, page.formData)).map(err => {
```
paulfgraham, line 14: The same thing with this interpolation of the errors - they're not _truly_ dynamic... but in this case, making them truly dynamic seems like overkill to me so I think this will do nicely.

Just flagging up that if you want an error message to change on the same page when a field label changes, this won't do it until you re-validate the page.

paulfgraham (commented):
A couple of things to draw your attention to... and that import would definitely benefit from being in the right spot and categorised correctly. 🙂

paulfgraham (approved):

PR #137, opened 2022-06-20
COP-10967: Fix for time input

paulfgraham:
### Description
Using version `"1.7.4"` of the component library, which includes the fixed `<TimeInput />` component. Also added some unit tests for the time input.

https://support.cop.homeoffice.gov.uk/browse/COP-10967

### To test
This is the broken behaviour, so make sure it's no longer happening in the Form Renderer Storybook:
1. Set up the JSON to have a time field;
2. Set up the initial data to have a valid value for that time field;
3. Switch to the Preview tab and you'll see the value appropriately showing in the time field;
4. Go back to the initial data and change to another valid value for the time field;
5. Switch back to the Preview tab and you'll now see the value in the time field flicking between the old and new values.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "post-compile": "rimraf dist/*.test.* dist/**/*.test.* dist/**/*.stories.* dist/docs dist/assets"
   },
   "dependencies": {
-    "@ukhomeoffice/cop-react-components": "1.7.3",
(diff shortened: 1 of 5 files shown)


georgepemberton1998:
I am still seeing the flickering behaviour when following the above steps

georgepemberton1998:
Looks good to me

georgepemberton1998 (approved):

PR #140, opened 2022-06-21
Fix bugs with option nested values

paulfgraham:
### Description
Fixed an issue with an undefined `page.formData` when showing the CYA screen (hub-and-spoke form type) when the form contains a `<Checkboxes />` component.

### To test
Added a unit test to make sure it now behaves itself.


gardnersam (approved):
Looks good

PR #144, opened 2022-06-27
BUG: Fix issue with page validation

paulfgraham:
### Description
A bug was introduced as part of interpolating error messages, which didn't account for collections and containers properly.

### Note
This shouldn't be merged with the pre-release suffix on the version.


keenan218 (approved):
Looks good 😄

PR #145, opened 2022-06-27
COP-10220: Improve support for refdata options

paulfgraham:
### Description
Added support for a custom item structure for reference data options that are returned, along with support for numeric identifiers that should be used as option values.

https://support.cop.homeoffice.gov.uk/browse/COP-10220

### To test
Covered by accompanying unit tests, but this will also be visible when setting up a form with differently structured reference data, such as the TIS ([COP-10220](https://support.cop.homeoffice.gov.uk/browse/COP-10220)).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "3.3.2",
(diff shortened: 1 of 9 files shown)


keenan218 (approved):
looks good

PR #153, opened 2022-07-06
EAB-569 Added additional validation methods for minimum and maximum length

LewisClark-HO:
Added extra `additionalValidation` methods for validating the length of inputs - `mustBeLongerThan` and `mustBeShorterThan` as per JIRA ticket: [EAB-659](https://support.cop.homeoffice.gov.uk/browse/EAB-659).

### To test
Covered by accompanying unit tests. Functionality can be tested in the Form Renderer Sandbox by adding the following JSON to a component:
```
...
"additionalValidation": [
  {
    "function": "mustBeLongerThan",
    "value": 3,
    "message": "Component value must be longer than 3 characters."
  }
]
...
```
Please note the `value` field is non-inclusive - the example above will only allow lengths of 4 or greater.


LewisClark-HO:
Currently investigating why my git insists on saying the whole file has changed when changing a few lines...

src/utils/Validate/additional/index.js
```diff
@@ -3,8 +3,14 @@ import mustBeAfter from './mustBeAfter';
 import mustBeBefore from './mustBeBefore';
 import mustBeInThePast from './mustBeInThePast';
 import mustBeInTheFuture from './mustBeInTheFuture';
+import mustBeLongerThan from './mustBeLongerThan';
+import mustBeShorterThan from './mustBeShorterThan';
 
-const functions = { mustBeAfter, mustBeBefore, mustBeInThePast, mustBeInTheFuture };
+const functions = { 
+  mustBeAfter, mustBeBefore,
```
paulfgraham, line 10: Can you list these out fully vertically, rather than in a 2 x 3 grid?

src/utils/Validate/additional/mustBeLongerThan.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Function returns true if the string parameter is longer than a given length, false
+ * if it is not.
+ * @param {string} string - the string to check the length of.
+ * @param {boolean} config.value - the value that string.length must be greater than.
```
paulfgraham, line 5: The param is an `object` that contains a `boolean` value... so this is incorrectly commented.

src/utils/Validate/additional/mustBeShorterThan.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Function returns true if the string parameter is shorter than a given length, false
+ * if it is not.
+ * @param {string} string - the string to check the length of.
+ * @param {boolean} config.value - the value that string.length must be less than.
```
paulfgraham, line 5: Again, the param is an `object` that contains a `boolean` value... so this is incorrectly commented.

paulfgraham (changes requested):

src/utils/Validate/additional/mustBeLongerThan.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Function returns true if the string parameter is longer than a given length, false
+ * if it is not.
+ * @param {string} string - the string to check the length of.
+ * @param {boolean} config.value - the value that string.length must be greater than. 
+ * @returns true if string.length is greater than config.value, false if not.
+ */
+const mustBeLongerThan = (string, config) => {
+  return (string.length > config.value);
```
paulfgraham, line 9: What about if `string` is `null` or `undefined`?

src/utils/Validate/additional/mustBeShorterThan.js
LewisClark-HO: What's the correct way to comment this - is it as `{object}` or as `{number}`? I'm not sure what's best.

src/utils/Validate/additional/mustBeLongerThan.js
```diff
@@ -0,0 +1,16 @@
...
+ * Function returns true if the string parameter is longer than a given length, false
+ * if it is not. An empty string will cause the function to return true - this is to  
+ * allow non-required fields to use this kind of validation.
+ * @param {string} string - the string to check the length of.
+ * @param {number} config.value - the length that string.length must be greater than. 
+ * @returns true if string.length is greater than config.value, false if not.
+ */
+const mustBeLongerThan = (string, config) => {
+  if (!string) {
+    return true;
```
paulfgraham, line 11: `undefined.length > 3 === true`? 🤨

src/utils/Validate/additional/mustBeLongerThan.test.js
```diff
@@ -0,0 +1,24 @@
...
+          const result = mustBeLongerThan('test', {value: 3});
+          expect(result).toEqual(true);
+        });
+
+        test('should return false given a string shorter than the given length', () => {
+          const result = mustBeLongerThan('to', {value: 3});
+          expect(result).toEqual(false);
+        });
+
+        test('should return false given a string equal to the given length', () => {
```
paulfgraham, line 17: There are no tests for when the text is `null` or `undefined`...

src/utils/Validate/additional/mustBeShorterThan.test.js
```diff
@@ -0,0 +1,24 @@
...
+          const result = mustBeShorterThan('to', {value: 3});
+          expect(result).toEqual(true);
+        });
+
+        test('should return false given a string longer than the given length', () => {
+          const result = mustBeShorterThan('test', {value: 3});
+          expect(result).toEqual(false);
+        });
+
+        test('should return false given a string equal to the given length', () => {
```
paulfgraham, line 17: There are no tests for when the text is `null` or `undefined`...

paulfgraham (changes requested):

src/utils/Validate/additional/mustBeLongerThan.js
paulfgraham: This remains unresolved and, actually, line 13 is plain wrong as it allows an empty string to be considered to have a length greater than specified:
`null.length > 3 === true`?
`''.length > 3 === true`?

src/utils/Validate/additional/mustBeLongerThan.test.js
```diff
@@ -0,0 +1,39 @@
...
+        });
+
+        test('should return false when string is undefined', () => {
+          const result = mustBeLongerThan(undefined, {value: 3});
+          expect(result).toEqual(false);
+        });
+
+        test('should return true when string is empty', () => {
+          const result = mustBeLongerThan('', {value: 3});
+          expect(result).toEqual(true);
```
paulfgraham, line 29: Yeah, why is this `true`? `''` is a string of length `0` (zero).

src/utils/Validate/additional/mustBeLongerThan.test.js
```diff
@@ -0,0 +1,39 @@
...
+        });
+
+        test('should return true when string is null', () => {
+          const result = mustBeLongerThan(null, {value: 3});
+          expect(result).toEqual(true);
```
paulfgraham, line 34: And why does this differ from `undefined`? That doesn't make sense to me?

src/utils/Validate/additional/mustBeShorterThan.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Function returns true if the string parameter is shorter than a given length, false
+ * if it is not.
+ * @param {string} string - the string to check the length of.
+ * @param {boolean} config.value - the value that string.length must be less than. 
+ * @returns true if string.length is less than config.value, false if not.
+ */
+ const mustBeShorterThan = (string, config) => {
+  return (string.length < config.value);
```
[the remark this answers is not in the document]
paulfgraham: Again, we have inconsistency here between the behaviour of `undefined`, `null`, and `''`.

I'd be inclined to always treat `undefined` and `null` the same as `''` in these methods - i.e., they should be considered to be strings of length zero.

You can either account for them directly, sort of as you're doing, or you can simply do something like this:
```
const str = string || '';
return str.length < config.value;
```

paulfgraham (changes requested):

src/utils/Validate/additional/mustBeLongerThan.test.js
LewisClark-HO: My thinking was string being `null` would have to be intentional, so it should pass but `undefined` could be caused by an error. 

What you say below about treating both as an empty string makes perfect sense though - thank you!

src/utils/Validate/additional/mustBeLongerThan.test.js
paulfgraham: Given this, I would suggest both should look like this:
```
const mustBeLongerThan = (string, config) => {
  if (!string) {
    // null, undefined and empty strings should be picked up by the required flag
    // and not considered here as they would be valid for optional fields.
    return true;
  }
  return string.length > config.value; // And the reverse for mustBeShorterThan
};
```

I think it's important to explain why an empty string, which is _not_ longer than the length specified is considered valid and, equally, why it's treated exactly the same way on the shorter validation.

LewisClark-HO: I think they are aren't they? There doesn't need to be an explicit catch for that in `mustBeShorterThan` because they will all pass by default unless someone sets value as something silly, i.e. <= 0.

LewisClark-HO: Yeah, good point! Will add the comments.

paulfgraham: > I think they are aren't they? There doesn't need to be an explicit catch for that in `mustBeShorterThan` because they will all pass by default unless someone sets value as something silly, i.e. <= 0.

Yes, they would all pass with your current implementation, but calling it out as being specifically ignored and considered valid is really useful from a code maintenance point of view. 🙂

PR #154, opened 2022-07-11
Cop-9749/view profile

keenan218:
Part of manage my profile. Added new cop-userProfile form inclusive of Location. 

Updated storybook with Location sections and corresponding refData.


src/json/cop-userProfile.json
```diff
@@ -0,0 +1,360 @@
+{
```
paulfgraham, line 1: Is this the one you intend to use for editing a user's profile? If so, it ought to be in the `cop-forms` repository, not in here. If not, and if this is simply for the purposes of getting it working, this shouldn't form part of the codebase - we really need to find an alternative to adding reference data to this repository as it exposes something that is otherwise secured behind an authentication layer.

keenan218: If that is the case i shall take it out then, wasn't a necessity to be here, r.e the form ill put that in the `cop-forms` repo.

PR #155, opened 2022-07-11
change paragraph font size from 16 to 19px

bowerscodes-COP:
### Description
Changed font size from 16px to 19px in Markup component

([link_to_jira_story](https://support.cop.homeoffice.gov.uk/browse/COP-11027))


aliktb (approved):
Looks good

PR #157, opened 2022-07-12
EAB-691 Fixed clicking an ErrorSummary link wiping nested components

LewisClark-HO:
The JIRA ticket for this issue is [EAB-691](https://support.cop.homeoffice.gov.uk/browse/EAB-691).

This is the same issue that was solved for regular components in [CI-130](https://support.cop.homeoffice.gov.uk/browse/CI-130). The previous fix didn't account for nested components.

The `formData` passed to a component is now merged with a component's `patch` data. This will replace any missing nested values if they are present in `patch`.


aliktb (approved):
Approved

PR #164, opened 2022-07-15
Added support for "contains" operator on conditional display

georgepemberton1998:
### Description
show_when can now be passed on a component or page with the `contains` operator. This is different to the `in` operator in that the in operator is used to pass an array of accepted values that a given field is "in", whereas `contains` is used to pass a single accepted value which is "in" a field.

### To test
Covered by accompanying unit tests. Functionality can be tested in the Form Renderer Sandbox by adding a component with a show_when with "op": "contains".
1. Add an array field to the `initial data` tab in form renderer sandbox. e.g. "something" : ["alpha", "bravo", "charlie" ]
2. use the below component and confirm that it is displayed
```
    {
      "id": "status",
      "fieldId": "status",
      "type": "text",
      "label": "select status",
      "required": true,
      "show_when": [
        {
          "field": "something",
          "op": "contains",
          "value": "alpha"
        }
      ]
    }
```
3. change the component check for a value not in the array and confirm that it is no longer displayed
```
    {
      "id": "status",
      "fieldId": "status",
      "type": "text",
      "label": "select status",
      "required": true,
      "show_when": [
        {
          "field": "something",
          "op": "contains",
          "value": "delta"
        }
      ]
    }
```

### DIFF
--- a/src/docs/ConditionalDisplayExamples.stories.mdx
+++ b/src/docs/ConditionalDisplayExamples.stories.mdx
@@ -173,7 +173,7 @@ In the above example, the `BMWModel` component **will** be shown when the value
 of the `cars` field **contains** `BMW`.
 
 In this scenario, the component will be shown if the `cars` field is one of the following: <br/>
-An array containing `BMW` e.g. `"cars": "['Ford', 'Mercedes', 'BMW', 'Lexus']"` <br/>
+An array containing `BMW` e.g. `"cars": ['Ford', 'Mercedes', 'BMW', 'Lexus']` <br/>
 A string containing `BMW` e.g. `"cars": "Ford,Mercedes,BMW,LEXUS"` <br/>
 Note: `contains` also works for integer arrays
 
--- a/src/docs/ConditionalDisplayExamples.stories.mdx
+++ b/src/docs/ConditionalDisplayExamples.stories.mdx
@@ -154,3 +154,28 @@ if the value of `areYouACivilServant` is "yes", `delegateEmails` will still be s
 ### When to use type `or`
 
 Use type `or` when you want a component, page or container to be shown if **at least one** of the conditions you provide are true.
+
+## Contains
+
+```json
+{
+  "id": "BMWModel",
+  /* ... */
+  "show_when": {
+    "field": "cars",
+    "op": "contains",
+    "values": "BMW"
+  }
+}
+```
+
+In the above example, the `BMWModel` component **will** be shown when the value
+of the `cars` field **contains** `BMW`.
+
+In this scenario, the component will be shown if the `cars` field is one of the following: <br/>
+An array containing `BMW` e.g. `"cars": "['Ford', 'Mercedes', 'BMW', 'Lexus']"` <br/>
+A string containing `BMW` e.g. `"cars": "Ford,Mercedes,BMW,LEXUS"` <br/>
+Note: `contains` also works for integer arrays
+
+### When to use a `contains` operator
+Use the `contains` operator when you want to show a component on the condition that a field contains a given value. 
--- a/src/utils/Condition/meetsCondition.js
+++ b/src/utils/Condition/meetsCondition.js
@@ -46,7 +46,11 @@ const meetsCondition = (condition, value) => {
         return true;
       }
       case 'contains': {
-        return value.toString().includes(compare);
+        if (value){
+          return value.toString().includes(compare);
+        }
+       // If no value is provided, the field cannot contain it, so it must fail the condition.
+       return false; 
       }
       default:
         return false;
--- a/src/utils/Condition/meetsCondition.js
+++ b/src/utils/Condition/meetsCondition.js
@@ -46,11 +46,8 @@ const meetsCondition = (condition, value) => {
         return true;
       }
       case 'contains': {
-        if (value){
-          return value.toString().includes(compare);
-        }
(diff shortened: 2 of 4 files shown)


georgepemberton1998:
Converting to a draft until PR #163 has been merged

src/utils/Condition/meetsCondition.js
```diff
@@ -45,6 +45,9 @@ const meetsCondition = (condition, value) => {
         // If it's not an array, nothing can be IN it, so it must meet the condition.
         return true;
       }
+      case 'contains': {
+        return value.toString().includes(compare);
```
paulfgraham, line 49: If `value` is `null` or `undefined`, this will fail. It's also case-sensitive, which _may_ be the desired behaviour but may actually just be something that hasn't been considered. If the former, can we have a note expressing why case sensitivity is important.

src/docs/ConditionalDisplayExamples.stories.mdx
```diff
@@ -154,3 +154,28 @@ if the value of `areYouACivilServant` is "yes", `delegateEmails` will still be s
...
+    "values": "BMW"
+  }
+}
+```
+
+In the above example, the `BMWModel` component **will** be shown when the value
+of the `cars` field **contains** `BMW`.
+
+In this scenario, the component will be shown if the `cars` field is one of the following: <br/>
+An array containing `BMW` e.g. `"cars": "['Ford', 'Mercedes', 'BMW', 'Lexus']"` <br/>
```
paulfgraham, line 176: You don't need this for arrays unless you want to consider sub-strings of the items in arrays. For arrays, `op: 'in'` does that job.

Also, in your example, `cars` is _not_ an array, it's a string representation of an array.

src/utils/Condition/meetsCondition.test.js
```diff
@@ -269,6 +269,69 @@ describe('utils.Condition.meetsCondition', () => {
...
+      const CONDITION = getCondition(op, VALUE)
+      expect(meetsCondition(CONDITION, FIELD)).toBeFalsy();
+    });
+    it('should reject any value when the field is an empty string', () => {
+      const FIELD = '';
+      const VALUE = 'alpha';
+      const CONDITION = getCondition(op, VALUE)
+      expect(meetsCondition(CONDITION, FIELD)).toBeFalsy();
+    });
+
```
paulfgraham, line 332: As mentioned in a previous comment, you need tests for `null` and `undefined` values.

paulfgraham (changes requested):

src/docs/ConditionalDisplayExamples.stories.mdx
georgepemberton1998: My understand with `op: 'in'` is that it can be used to provide an `array value` and check if a `field` is "in" that array value. The `contains` operator would essentially do the opposite of this. You can provide a `field array` and check if a given value is in that. I looked at uplifting the "in" operator to enable this functionality but I believed it to be quite confusing.

To clarify - as part of the context data we have an array field "groups" (below) - I would like to display a component if `"/bf-intel"` is in the `groups` array. I was unable to see a way to use this with the "in" operator without a code change to it. Please let me know if I have missed something.

```
"shiftDetailsContext": {
    [
     ---
    "groups" [
      "/bf-intel",
      "/bfrole",
      "/camunda-admin"
    ]
    ---
    ]
}
```

paulfgraham: Ah, I see what you mean. Yes, you're absolutely right!

The point about it being a string representation of an array remains, though. 🙂

src/utils/Condition/meetsCondition.js
```diff
@@ -46,7 +46,11 @@ const meetsCondition = (condition, value) => {
         return true;
       }
       case 'contains': {
-        return value.toString().includes(compare);
+        if (value){
```
paulfgraham, line 49:
```suggestion
        if (value) {
```

Alternatively, this whole thing could simply be:
```
        return value?.toString().includes(compare);
```

src/utils/Condition/meetsCondition.js
```diff
@@ -46,7 +46,11 @@ const meetsCondition = (condition, value) => {
...
+          return value.toString().includes(compare);
+        }
+       // If no value is provided, the field cannot contain it, so it must fail the condition.
```
paulfgraham, line 52: The indenting on this line and the one below is off - it's a single space too far back.

src/utils/Condition/meetsCondition.test.js
```diff
@@ -329,6 +329,30 @@ describe('utils.Condition.meetsCondition', () => {
...
+      const CONDITION = getCondition(op, VALUE)
+      expect(meetsCondition(CONDITION, FIELD)).toBeFalsy();
+    });
+    it('should reject any field when the value is undefined', () => {
+      const FIELD = ['alpha','bravo','charlie'];
+      const VALUE = undefined;
+      const CONDITION = getCondition(op, VALUE)
+      expect(meetsCondition(CONDITION, FIELD)).toBeFalsy();
+    });
+    it('should reject any value when the field is undefined', () => {
```
paulfgraham, line 344:
```suggestion
    it('should reject any value when the field is null', () => {
```

paulfgraham (changes requested):

PR #167, opened 2022-07-20
Update cop-react-components to version 1.10.0

andrewgibson-ibm:
Following approval and merge of https://github.com/UKHomeOffice/cop-react-components/pull/108


gardnersam (changes requested):
There are too many changes to the yarn.lock. It should just change a few lines to update the cop-react-components version. 

I think maybe if you do the npm install of the latest version it can mess with the yarn.lock. So the order might matter, or potentially you could try doing a npm install, then resetting yarn.lock then yarn adding.

PR #171, opened 2022-07-26
Filter select values based on value selected in another select on the…

Deepak-Bagrawat:
Filter the select values displayed in the drop-down list  based on the value selected in a different select (https://support.cop.homeoffice.gov.uk/browse/EAB-581)


src/utils/Data/setupRefDataUrlForComponent.js
```diff
@@ -24,6 +24,9 @@ const setupRefDataUrlForComponent = (component, data) => {
     if (component.type === ComponentTypes.COLLECTION) {
       return setupRefDataForCollection(component, data);
     }
+    if (component.type === ComponentTypes.SELECT) {
+      return component;
```
paulfgraham, line 28: Why is this here? This would mean that a select cannot have its values from reference data...?

paulfgraham: Thanks for the response, @Deepak-Bagrawat. That makes a lot of sense.

paulfgraham (approved):

PR #173, opened 2022-07-27
COP-11361 - Render entry point to a collections on CYA page

charles474:
This PR adds the capability of the CYA page rendering a row which is an entry point to a collection page when a collection has no items in it.

This will allow a user to add and remove items from a collection at will from the CYA screen.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "3.10.0",
+  "version": "3.11.1",
```
paulfgraham, line 3: Why has this jumped by both minor and patch?

charles474: Minor as it wasn't exactly a bug fix unless i've understood it wrong and patch was due to wording change i had between last pre-release I did. More than happy to be corrected 😉

paulfgraham: A minor increment should have a patch number of `0`. I'm slightly confused why it's increased... if you did pre-releases as, say, `3.11.0-alpha` and `3.11.0-beta`, the subsequent release should be `3.11.0`. If you've accidentally gone up to `3.11.1-something`, fair enough... but the patch number shouldn't be changing in pre-releases, just the suffix. 🙂

charles474: Right gotcha 👍

charles474: Have corrected 😄

paulfgraham (approved):

PR #175, opened 2022-07-31
WIP - Adds onCancel hook

charles474:
This PR allows a cancel action to be given to the form renderer as a hook.

By default it does nothing, the function to action **MUST** be passed from the consuming application.


src/hooks/useHooks.test.js
```diff
@@ -108,6 +108,20 @@ describe('hooks', () => {
       expect(typeof hooks[HOOK_NAME]).toEqual('function');
     });
 
+    it('should call onCancel function', async () => {
```
paulfgraham, line 111: This will be covered by the `ALLOWED_HOOKS.forEach(...)` thing above, won't it? I'm not sure what this test is adding... 🤔

paulfgraham (changes requested):
Just a formatting tweak, with an inline suggestion.

src/hooks/useHooks.test.js
charles474: No you are right, i think i was just trying to be fancy in adding another test that was not needed.

src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -171,6 +171,10 @@ const InternalFormRenderer = ({
 
   // Handle actions from pages.
   const onPageAction = (action, patch) => {
+    if(action.type === PageAction.TYPES.CANCEL){
```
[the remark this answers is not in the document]
charles474: Bloody linting 😁

paulfgraham (approved):

PR #178, opened 2022-08-09
BFCMS-254: Handle page action completing task

paulfgraham:
### Description
Added code to use a `complete` flag on a page action in the context of tasks in a task list. This removes the hard requirement to go through a Check your answers page to complete a task, although that remains the default approach.

### To test
Covered by an accompanying unit test. You can alternatively try it out in the Sandbox by setting up a task list form and switching out the final `saveAndNavigate` action for a `save` action with `complete: true` as one of its properties. All of the other action properties (such as `validate: true`) should still be respected.


gardnersam (approved):
Looks good, thanks for the tidy up as well !

paulfgraham:
> Looks good, thanks for the tidy up as well !

Ha! I forgot I'd actually done that. 😂

PR #179, opened 2022-08-09
BFCMS-255: Support completed tasks not showing CYA

paulfgraham:
### Description
Alongside the ability to complete a task without the need for a Check your answers page, it makes no sense to show the Check your answers page when you subsequently revisit that completed task. This adds support for returning the user to the page on which they marked the task as "Completed", which is the circumstance in which NOT having a Check your answers page makes the most sense.

### To test
Covered by the accompany additional unit test. To try this out in the sandbox, set `hub.noTaskCYAs` to `true` in the Form JSON.


gardnersam (approved):
Looks good to me

PR #180, opened 2022-08-09
BFCMS-256: Conditionally show buttons

paulfgraham:
### Description
Added support for conditionally showing buttons, leverage the exact same mechanism as for components and pages - namely, the button will be shown if its `show_when` conditions are met; otherwise, it will be hidden.

### To test
Covered by an accompanying unit test. You can also test this in the sandbox by setting up a `show_when` on a button and seeing it show or hide as you change the contents of the page on which that button appears.


gardnersam (approved):
Looks good

PR #181, opened 2022-08-09
Default task state

paulfgraham:
### Description
The form JSON configuration shouldn't be specifying the state of each task as that's dynamic and comes from the form _data_, not its configuration. At the moment, you can't set a form up without specifying the state of each task or else you get console errors thrown... so defaulting it to the state it automatically sets them to anyway (Cannot start yet) seems sensible.

Also fixed an indenting issue in the `getComponent` utility function, which had been bugging me for a while.

### To test
The task state defaulting is covered by accompanying unit tests. There is functionally no difference with the current behaviour but you can see console errors right now without `state` specified on each and every task in the Form JSON whereas you won't get those errors once this PR is merged and released.


charles474 (approved):
LGTM

PR #184, opened 2022-08-10
COP-11597 - Dependency bump, version bump

charles474:
This PR bumps the `cop-react-components` dependency and bumps the `cop-react-form-renderer` version number.


paulfgraham (approved):
It looks like the adjustment to the tests is completely reasonable.

PR #185, opened 2022-08-11
BFCMS-238: Add support for non-sequential tasks

paulfgraham:
### Description
Added support for non-sequential tasks in the task list. This results in all tasks having a default state of "Not started" unless it depends on other sections having been completed (via a `depends_on` property in the form JSON) - in this circumstance, it will show as "Cannot start yet" until such time as all of the sections it depends on have been completed, whereupon it will show as "Not started".

### To test
Covered by accompanying unit tests. You can see this behaviour in the sandbox by adding `nonSequential: true` to the `hub` in the Form JSON and then setting up one of the tasks to have `depends_on: [ "Some other task name" ]`.


gardnersam (approved):
Looks good to me

PR #187, opened 2022-08-12
Allow no page actions

paulfgraham:
### Description
A previous PR for conditionally showing buttons introduced a bug where page actions became mandatory because of the way the code was written. This returns that to its previous situation where a page can quite happily have no actions.


gardnersam (approved):
looks good to me

PR #195, opened 2022-08-24
Fix bug where CYA doesn't work with text above collection on a page

gardnersam:
If you have a page with text followed by a collection component then the CYA screen breaks as the text component has no label. This PR fixes that.


LewisClark-HO (approved):
LGTM

PR #198, opened 2022-08-25
Cop-11552/cya-keyboard-navigation

keenan218:
Linking to bug [cop-11552](https://support.cop.homeoffice.gov.uk/browse/COP-11552)

Allowed user to navigate through a cya page using tab and access any change links through the enter key


aliktb (approved):
Looks good

src/components/SummaryList/RowAction.jsx
```diff
@@ -17,7 +17,14 @@ const RowAction = ({ row }) => {
   }
 
   return (
-    <Link {...attrs}>
+    <Link {...attrs}
+      onKeyPress={(event) => {
```
paulfgraham, line 21: @keenan218: Any reason why this wasn't added to `getRowActionAttributes`?

keenan218: I didn't consider it if i'm honest, i found the place where the tabIndex worked correctly and just assumed it would all go in there. Would it be better suited in the return with the onClick?

paulfgraham: It would keep all that logic in one place, which I think makes it cleaner. Also, this mechanism requires that there is a `row.action.onAction` method or it'll throw an exception - the `getRowActionAttributes` checks for the presence of `row.action.onAction` before deciding whether to include an `onClick` (and potentially now an `onKeyPress`) or ` simple `href`.

keenan218: Okay i see, would you suggest the `onKeyPress` needs to be separate from the `onClick`? also the `event` on line 21 works in `RowAction` but putting it into `getRowActionAttributes` event doesn't seem to work in .js

paulfgraham: It should be distinct from the `onClick`, yes. But are you saying this doesn't work?
```
const getRowActionAttributes = (row) => {
  if (row && row.action) {
    if (typeof(row.action.onAction) === 'function') {
      return {
        onClick: () => row.action.onAction(row),
        onKeyPress: (event) => {
          if (event.key === 'Enter') {
            row.action.onAction(row);
          }
        },
        tabIndex: '0'
      };
    } else if (row.action.page) {
      return { href: `/${row.action.page}`};
    }
  }
  return {};
};

export default getRowActionAttributes;
```

keenan218: Ah yes, i had it slightly different.. That works! I agree does also look cleaner.

PR #204, opened 2022-08-31
EAB-604 add container group validation

gardnersam:
Expands container validation to allow running custom validators on a whole container before validating the contained components. 

Adds first custom validator `mustEnterAtLeastOne` that ensures the user has filled in at least one component within the container. 

The container is wrapped in a FormGroup if custom validators are present to allow displaying the error over the whole container. This allows functionality like this:

image


src/utils/Validate/additional/index.js
```diff
@@ -5,20 +5,22 @@ import mustBeInThePast from './mustBeInThePast';
...
   mustBeInThePast, 
   mustBeInTheFuture, 
   mustBeLongerThan, 
-  mustBeShorterThan 
+  mustBeShorterThan,
+  mustEnterAtLeastOne
 };
 
-const additionalValidation = (value, config) => {
+const additionalValidation = (value, config, component) => {
```
LewisClark-HO, line 20: Just wondering if the extra argument needs adding - can't you access the collection's components from `config`?

Deepak-Bagrawat (approved):

src/utils/Validate/additional/index.js
gardnersam: No unfortunately, it's badly named. it's defined as config on line 31 and is actually an entry in the additionalValidation array, rather than the components config as I also assumed at first

LewisClark-HO: Ahh ok, that makes sense! Thank you. Approved :)

PR #206, opened 2022-09-02
Cop 11808/back link

fisherrj2:
Fixes a bug when going back after changing data on a page. Although the CYA page data looked up to date, the submission data still reflected the status prior to going back. Fix is to reset submission data in pop state handler.  See https://support.cop.homeoffice.gov.uk/browse/COP-11808


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "3.25.0",
+  "version": "3.25.1",
```
aliktb, line 3: Granted it is a bug but I would set it as `3.26.0` just in case since it is some new changes we are adding in

aliktb (commented):
Looks fine. Just one comment on the version

package.json
fisherrj2: Changed as advised

aliktb (approved):

PR #208, opened 2022-09-02
Fix-for-unable-to-load-refData-in-nested=component

Deepak-Bagrawat:
https://support.cop.homeoffice.gov.uk/browse/EAB-991


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "3.26.0",
+  "version": "3.25.1",
```
gardnersam, line 3: it's already at 3.26. It can't go down to 25. This should be 2.26.1

PR #219, opened 2022-09-08
Feature/eab 987 make actions disablable

gardnersam:
Two things in one, seemed small enough to combine.

Firstly this allows you to disable an action button if `disabled` is set on that action.

Secondly it allows you to specify some form data you'd like to create / edit for the current page.


LewisClark-HO (approved):
LGTM

PR #232, opened 2022-09-20
EAB-950 Add actions for page collections

gardnersam:
Add change and remove actions for page collections.


src/utils/CheckYourAnswers/getCYACollectionChangeAction.js
```diff
@@ -0,0 +1,30 @@
...
+
+const getCYACollectionChangeAction = (page, item, onAction, labelCount) => {
+  const labels = {
+    ...CollectionLabels,
+    ...page.collection.labels,
+  };
+  return {
+    pageId: page.id,
+    fieldId: item.id,
+    key: item.id,
```
LewisClark-HO, line 15: Will this not create rows with duplicate keys? If all actions are using `item.id` as the key.

src/utils/CheckYourAnswers/getCYACollectionChangeAction.js
```diff
@@ -0,0 +1,32 @@
...
+  return {
+    pageId: page.id,
+    fieldId: `change${labelCount}${action.page}`,
+    key: `change${labelCount}${action.page}`,
+    type: 'action',
+    action: {
+      onAction: (row) => {
+        const changeCallback = (data) => {
+          data[`${page.collection.name}ActiveId`] = item.id;
+          data.currentCollection = page.collection.name;
```
LewisClark-HO, line 18: What is `currentCollection` used for? My bad if it's something I added, can't remember what it's doing...

gardnersam: It's new, and the purpose is so that if we then navigate to another page where the delete action will be used, we need to know what collection to delete from. Otherwise we just have an activeId and potentially several collections. I guess we could filter on activeId through all the collections. I don't think its obvious programatically to tell what in the form data is a collection however

src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -225,6 +225,10 @@ const InternalFormRenderer = ({
         }
         // Save draft or submit.
         const submissionData = Utils.Format.form({ pages, components }, { ...data, ...patch, ...formState.page.formData }, EventTypes.SUBMIT);
+        if(action.secondaryType === 'deleteCollectionEntry'){
```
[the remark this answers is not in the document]
gardnersam: As in add a page action like PageAction.TYPES.COLLECTION_DELETE: ?

src/utils/CheckYourAnswers/getCYACollectionDeleteAction.js
```diff
@@ -0,0 +1,27 @@
...
+const getCYACollectionDeleteAction = (page, item, onAction, labelCount, action) => {
+  const label = action?.label || CollectionLabels.remove;
+  return {
+    pageId: page.id,
+    fieldId: `delete${labelCount}`,
+    key: `delete${labelCount}`,
+    type: 'action',
+    action: {
+      onAction: () => {
+        const changeCallback = (data) => {
```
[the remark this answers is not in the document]
gardnersam: Good shout

src/components/FormRenderer/FormRenderer.jsx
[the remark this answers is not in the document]
LewisClark-HO: Yeah sorry, should've made that clearer. It will have to be a page action anyway for the confirmation pages I think.

src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -225,6 +225,10 @@ const InternalFormRenderer = ({
...
+          const activeId = submissionData[`${submissionData.currentCollection}ActiveId`];
```
LewisClark-HO, line 229: Just seen this is where `currentCollection` is used. In this context, you can just use `action.collection` to get the name of the collection that the action was taken on. No need to create a dedicated field for it I don't think.

gardnersam: Thanks for pointing that out, I've removed currentCollection

src/components/FormRenderer/FormRenderer.jsx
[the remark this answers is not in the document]
gardnersam: I've swapped this logic into a page action, good plan

LewisClark-HO (approved):
Approved, thanks for updating the storybook too!

PR #235, opened 2022-09-22
Support getting data from top level refdata response

LewisClark-HO:
### Description
Added support for retrieving values from the top-level of a refData response if they are not found under `data`. 

### Testing
Associated unit tests have been added.

Not sure why yarn.lock decided to rearrange itself.


gardnersam (approved):
LGTM

PR #242, opened 2022-10-05
Added checkValueIsTruthy page operation

LewisClark-HO:
### Description
Added a new page operation for checking if a value in a page's `formData` is truthy or not. Two new conditions have also been added to the `run_when` clause of a page operation so that operations can be run depending on another value's truthy-ness.

### Examples
The sandbox documentation has been updated with an example of this new operation.

### Testing
Accompanying unit tests are included.


src/docs/PageOperationExamples.stories.mdx
```diff
@@ -113,6 +150,7 @@ Operations can be chained, with the results of earlier operations being used in
       "output": "retrievedRegistration"
     }
   ]
+  ...
 }
 ```
 In the example above, the fist operation will find an entry in the `cars` array with a `model` that matches
```
mark-roberts-ho, line 156: Can you correct the spelling of 'first'?

src/docs/PageOperationExamples.stories.mdx
```diff
@@ -139,8 +178,56 @@ Operations can be set to only run once a certain condition has been met. This is
...
+```json
+{
+  "id": "examplePage",
+  "name": "examplePage",
+  "title": "Example page",
+  ...
+  "operations": [
+    {
+      "run_when": {
+        "condition": "isFalsey",
```
mark-roberts-ho, line 221: I realise `falsey` is legitimate but it's used far less often than `falsy`

mark-roberts-ho (commented):
Just minor quibbles

PR #244, opened 2022-10-07
Updated return type of operation

LewisClark-HO:
### Description
Updated return type of the `getIndexOfMatchingValueIn` page operation to return a string rather than number. This change was made because of the way integer 0 was not working with `interpolateString`.

This PR also updates the version of `cop-react-components` being used to 2.1.0.


package.json
```diff
@@ -16,7 +16,7 @@
     "post-compile": "rimraf dist/*.test.* dist/**/*.test.* dist/**/*.stories.* dist/docs dist/assets"
   },
   "dependencies": {
-    "@ukhomeoffice/cop-react-components": "2.0.3",
+    "@ukhomeoffice/cop-react-components": "^2.1.0",
```
gardnersam, line 19: not sure if we should have the ^

gardnersam (approved):

PR #246, opened 2022-10-11
Added support for operations to use a dot-separated fieldId for their output field

LewisClark-HO:
### Description
This PR adds support for page operations being able to output to a dot-separated field ID within `formData`.

### Examples
The storybook has been been updated with relevant examples.

### Testing
Accompanying unit tests are included.


src/utils/Operate/runPageOperations.js
```diff
@@ -19,29 +23,35 @@ const doOperation = (config, data) => {
...
-        if (shouldRun(op.run_when, data)) {
-          result[op.output] = doOperation(op, { ...data, ...result });
+      if (shouldRun(op.run_when, data)) {
+        const result = doOperation(op, data); 
+        if (op.output) {
+          let outputPath = Utils.interpolateString(op.output, data);
+          // Remove array indexers and replace them with dot
+          // separators. This ensures that outputPath will
+          // work with setDataItem.
+          outputPath = outputPath.replace('[', '.');
```
mark-roberts-ho, line 47: I don't like repeated assignments to same value:

Do replacement as one operation with chaining:

`outputPath = outputPath.replace( '[', '.').replace( ']', '')`

In fact could also chain it with the declaration
`let outputPath = Utils.interpolateString(op.output, data).replace(...).replace(...);`

I think that means outputPath can be declared as `const` as a result

mark-roberts-ho (changes requested):
See review comments about outputPath

mark-roberts-ho (approved):

PR #247, opened 2022-10-12
EAB-1130 enable collections within collections

gardnersam:
Fixes two problems with collections within collections. Firstly the CollectionPages view of the form data wasn't being updated properly, and secondly CYA screens were not prepared to handle a collection within a container, but they now are.


LewisClark-HO (approved):
LGTM

mark-roberts-ho (approved):
Approved but I think your let data could possibly be handled better by a setState or useEffect or something.

PR #251, opened 2022-10-13
Cop 8193/your profile

fisherrj2:
Changes to form renderer to allow for a 'hub-and-spoke' type form for manage my profile in cop-ui.  Support for going back within the form pages and also for checking the patch for changes and changing state on submission to show an alert panel.  See https://support.cop.homeoffice.gov.uk/browse/COP-8193


src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -206,7 +215,7 @@ const InternalFormRenderer = ({
         let pageUpdate = (next) => onPageChange(helpers.getNextPageId(type, pages, pageId, action, next));
         switch (action.type) {
           case PageAction.TYPES.SUBMIT:
-            setPagePoint('end');
+            setPagePoint('submit');
```
keenan218, line 218: Why has this been changed?

src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -98,7 +99,10 @@ const InternalFormRenderer = ({
 
   // Need to set submission data when going back
   window.onpopstate = () => {
-    if (components && pages && data && formState.page?.formData && pagePoint===undefined){
+    setGoingBack(true);
+    hooks.onGoingBack();
+    clearErrors();
```
keenan218, line 104: Would we need to clear errors here?

src/components/FormRenderer/FormRenderer.jsx
fisherrj2: It is just a better way of describing where we are in the page as submit is about to happen and depending on the pagePoint state there are certain actions such as setting data that need to happen as side effects depending on this state variable.  'submit' is also already used elsewhere in the file (see onCYAAction handler).

src/components/FormRenderer/FormRenderer.jsx
fisherrj2: If we don't clear errors here then if there has been a validation error on the current page, when going back to a previous page you will still see the error alert even though it doesn't apply to that page.

keenan218: I see, thanks

src/components/FormRenderer/FormRenderer.jsx
keenan218: Ah yes, that does make more sense

keenan218 (approved):
Looks good

PR #260, opened 2022-10-14
BFCMS-450: Support readonly containers and collections

paulfgraham:
### Description
Adding support for the `readonly` flag on collections and containers. This means that child components (including within collection items) will be readonly if the overall container/collection is marked as readonly.

https://jira.bics-collaboration.homeoffice.gov.uk/browse/BFCMS-450

### To test
Covered by additional accompanying unit tests.


src/components/FormComponent/Container.jsx
```diff
@@ -45,7 +45,10 @@ const Container = ({
         return <FormComponent
           key={index}
           {...attrs}
-          component={component}
+          component={{
+            ...component,
+            readonly: container.readonly ? true : component.readonly
```
LewisClark-HO, line 50: Purely personal preference but would `container.readonly || component.readonly` be more concise here?

LewisClark-HO (approved):
LGTM, one comment that's purely down to personal preference I think.

src/components/FormComponent/Container.jsx
LewisClark-HO: Ahh I see, that makes sense :)

PR #261, opened 2022-10-14
BFCMS-442: Support use within container

paulfgraham:
### Description
Pages in a form can display a field from the top-level array of `components` by way of a `use` reference. Previously, this could only reference the component directly but there is a need to reference components that sit _within_ a container component.

This functionality is introduced here and uses dot notation, which is consistent with the rest of the library.

https://jira.bics-collaboration.homeoffice.gov.uk/browse/BFCMS-442

### Example
```
{
  ...
  components: [
    {
      id: 'alpha',
      ...
      type: 'container',
      components: [
        id: 'bravo'
        ...
      ]
    }
  ],
  pages: [
    {
      ...
      components: [
        { use: 'alpha.bravo' }
      ]
    }
  ]
}
```

What this "use" displays is the specifically referenced component wrapped inside its container, but with none of its siblings, which means it remains fully nested and all functionality is maintained correctly.

### To test
Covered by additional accompanying unit tests.
You can also try it out in the Sandbox, following the above example and noting the ellipsis where you will need to populate those elements yourself.


michaelruocco (approved):
LGTM

charles474 (approved):
LGTM

PR #263, opened 2022-10-18
COP-12161 Rework tasklist flow

bowerscodes-COP:
Amend Tasklist flow to display CYA when a user saves & returns later, per [COP-12161](https://support.cop.homeoffice.gov.uk/browse/COP-12161)


src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -298,7 +298,7 @@ const InternalFormRenderer = ({
         }
       } else if (currentTask.state === TaskStates.TYPES.IN_PROGRESS) {
         const currentPage = data.formStatus.tasks[currentTask.name].currentPage;
-        onPageChange(currentPage || currentTask.pages[0]);
+        onPageChange(FormPages.CYA || currentTask.pages[0]);
```
fisherrj2, line 301: Just wondering if there is a need for the 'or' statement as in is it ever likely that FormPages.CYA would be null in this case?

PR #264, opened 2022-10-19
EAB-721: create sealNumberUnique validation method

mark-roberts-ho:
## Description

Adds a  method which when supplied with a seal number string returns true if the number already exists.
This method does **not** detect other errors such as whether the new seal number is valid


src/utils/Validate/additional/sealNumberUnique.js
```diff
@@ -0,0 +1,23 @@
...
+const sealNumberUnique = async (string) => {
+  const {data} = await axios.get(`${copServiceUrl}/epms/sealNumber/${string}`).catch(() => false)
+
+  if (data) {
+    // data.result is false is seal number exists, negate because we want true if error
+    return !data.result;
+  }
+
+  // if no data returned, assume seal number valid
+  return false
```
gardnersam, line 20: would this mean false is also returned in the condition where an error is caught and data is set to false?

PR #268, opened 2022-10-21
BFCMS-463: Use onChange handler in calculation component

paulfgraham:
### Description
The `config` passed to the `getCalculation` method includes an `onChange` function, which is how all other component types report any changes. This automatically plays nicely with containers and collections, whereas directly applying the value to the `formData` (also without considering nesting) does not.

Also using version 2.2.4 of the component library, which includes a fix for the prefix and suffix on `TextInput` components in a readonly mode - i.e., exactly what the calculation component uses for rendering.

https://jira.bics-collaboration.homeoffice.gov.uk/browse/BFCMS-463

### To test
Covered by updated accompanying unit tests.


charles474 (approved):
LGTM

PR #270, opened 2022-10-24
Cop 11899: Develop mandec forms in task list

andrewgibson-ibm:
# Description
Mandatory declarations has a requirement for it to be in a task list format; this is achieved by setting the `form` type to `task-list`
The required format needs pages _before_ the task-list hub is displayed to the user; this is achieved by setting the `form` type to `form-with-task-list`

## Problem
The current state of `form-with-task-list` lacks the following functionality:

1. Display labels for tasks on the task-list hub
2. Set the first page that a task navigates to when selecting a task from the task-list hub
3. If the `pre-task-list` pages are also pages used by a task, they should be set as `complete` once the user reaches the task-list hub

## Solution

1. Labels:
    1. Update `TaskList.jsx` to enable a `label` property in the form of a `Hint`
    2. Update `TaskList.stories.mdx` to reflect label
    3. Update `TaskList.test.js` to reflect label
2. First Page:
    1. Update `FormRenderer.jsx` to update the `currentPage` if `firstPage` has been set as a property in the `task`
    2. Update `taskList.json` to set a `firstPage` property
    3. Add test to `FormRenderer.test.js` to check that the `firstPage` is navigated to upon beginning a `task`
3. Complete Tasks:
    1. Update `getSubmissionStatus.js` to `complete` tasks based on if the `pre-task-list` page is part of a `task`
    2. Add test to `getSubmissionStatus.test.js` to check that a `task` is `complete`
4. Other:
    1. Update version in `package.json`

# Testing
- Units tests have been written as described above
- `yarn storybook` displays labels in the `task-list` story
- More integrated testing was done by:
  1. Creating a `form` with type set to `form-with-task-list`
      - [eab2 form](https://gitlab.digital.homeoffice.gov.uk/cop/cop-forms/-/blob/main/json/eab-add-a-border-event.json)
      - [mandec form](https://gitlab.digital.homeoffice.gov.uk/cop/cop-forms/-/blob/COP-11417-Create-mandec-forms-in-task-list-template/json/cop-make-your-mandatory-declarations.json)
  3. Deploying a `.bpmn` in `cop-local` which utilises the form
  4. Starting the process to complete the form
  5. Observing that the changes are successful

# Links
[Link to prototype](https://cop.prototype.cop.homeoffice.gov.uk/orange/mandec-current/task-list)
[Link to COP-11899](https://support.cop.homeoffice.gov.uk/browse/COP-11899)


src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -291,13 +291,13 @@ const InternalFormRenderer = ({
...
-          const currentPage = data.formStatus.tasks[currentTask.name].currentPage;
+          const currentPage = currentTask.firstPage ? currentTask.firstPage : data.formStatus.tasks[currentTask.name].currentPage;
           onPageChange(currentPage || currentTask.pages[0]);
         } else {
           onPageChange(FormPages.CYA);
         }
-      } else if (currentTask.state === TaskStates.TYPES.IN_PROGRESS) {
-        const currentPage = data.formStatus.tasks[currentTask.name].currentPage;
+      } else if (currentTask.state === TaskStates.TYPES.IN_PROGRESS || currentTask.firstPage) {
+        const currentPage = currentTask.firstPage ? currentTask.firstPage : data.formStatus.tasks[currentTask.name].currentPage;
```
gardnersam, line 300: same here, if a user save-and-returns on a page within a task they should go back to that page when they reenter the task. This looks like it will always take you to the first page

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ukhomeoffice/cop-react-form-renderer",
-  "version": "4.13.0",
+  "version": "4.16.1",
```
gardnersam, line 3: this seems like a big jump

src/components/FormRenderer/helpers/getSubmissionStatus.js
```diff
@@ -1,14 +1,40 @@
...
+        section.tasks.forEach((task) => {
+          let lastPage = task.pages[task.pages.length - 1]
+          if (((lastPage === currentPageId) || (lastPage.id === currentPageId)) && isCompleted) {
+            formStatus.tasks[task.name] = {
+              complete: isCompleted,
+              currentPage: currentPageId
+            };
+          } else if (task.pages.includes(currentPageId)) {
+            formStatus.tasks[task.name] = {
+              complete: false,
```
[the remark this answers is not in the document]
andrewgibson-ibm: - `currentTask` is set in `FormRenderer` -> `onTaskAction` -> `setCurrentTask`
- This method isn't called though until the user has selected a `task` from the hub screen
- When the user is in the `pre-task-list` pages, the `currentTask.name` is therefore `undefined`
- The logic I've added here `else if (!currentTask.name) {` is only entered when the user is in those `pre-task-list` pages
- We have pages we want the user to begin on but they are also the pages for the first two sections
- This means that with the current setup, the user would start the form, complete the first pages, then be asked again to complete the same pages because they are part of the first two sections
- This update means that if the user has completed them in the `undefined` `pre-task-list` part of the form then the renderer detects that and sets those sections to `complete`

src/components/FormRenderer/helpers/getSubmissionStatus.js
```diff
@@ -1,14 +1,40 @@
...
     formStatus.tasks[currentTask.name] = formStatus.tasks[currentTask.name] || {};
 
+
     if (currentPageId === (FormPages.CYA || 'submitForm') && isCompleted) {
       formStatus.tasks[currentTask.name].complete = true;
+    } else if (!currentTask.name) {
+      // undefined means that the currentTask isn't set because onTaskAction hasn't been called
+      // so the user is in the pre-task-list section of the form
+      // in this case check whether the page is part of a section defined in the hub sections
+      // if it is, set the section to complete as long as isCompleted == true
```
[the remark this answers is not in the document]
andrewgibson-ibm: Yeah as commented on the other conversation, it's that we want to be able to complete tasks from the initial pages before the task-list, otherwise the user has to fill them out twice

package.json
andrewgibson-ibm: [The current version in main is 4.16.0](https://github.com/UKHomeOffice/cop-react-form-renderer/blob/main/package.json#L3)

src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -291,13 +291,13 @@ const InternalFormRenderer = ({
       setCurrentTask(currentTask);
       if (currentTask.state === TaskStates.TYPES.COMPLETE) {
         if (hubDetails?.noTaskCYAs) {
-          const currentPage = data.formStatus.tasks[currentTask.name].currentPage;
+          const currentPage = currentTask.firstPage ? currentTask.firstPage : data.formStatus.tasks[currentTask.name].currentPage;
```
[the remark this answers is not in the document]
andrewgibson-ibm: Have just confirmed this so updated in latest commit
Tested to prove that user can both have `firstPage` set _and_ Save and return later now

src/components/FormRenderer/FormRenderer.jsx
andrewgibson-ibm: Have just confirmed this so updated in latest commit
Tested to prove that user can both have `firstPage` set _and_ Save and return later now

gardnersam (approved):
Looks good!

PR #271, opened 2022-10-24
COP-12195 - Validation error of nested container components

charles474:
# Description
This PR displays validation errors of components, nested within an already nested container, within another container (sample of JSON below), including collections.

e.g. 
```
{
  "type": "container",
  "components": [
    {
      "type": "text"
    },
    {
      "type": "container",
      "components": [
        {
          "type": "text"
        },
        {
          "type": "checkboxes"
        },
        {
          "type": "container",
          "components": [
            {
              "type": "checkboxes"
            },
            {
              "type": "container",
              "components": [
                {
                  "type": "radios"
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

### Supplemental
- This PR also addresses a bug where the container was unable to correctly set up the `full_path` for nested containers within collections.
- Fixes storybook deploy action


src/utils/Validate/validateContainer.js
```diff
@@ -25,10 +26,11 @@ const validateContainer = (container, outerData, formData) => {
   if (Array.isArray(container.components)) {
     const containerData = outerData && container.fieldId ? outerData[container.fieldId] : outerData;
     container.components.forEach(component => {
-      errors.push(validateComponent(component, containerData, fd));
+      const full_path = `${container.full_path || container.fieldId}.${component.fieldId}`;
+      errors.push(validateComponent({ ...component, full_path }, containerData, fd));
     });
   }
-  return errors.filter(e => !!e);
+  return elevateErrors(errors.filter(e => !!e));
```
LewisClark-HO, line 33: I'm not sure the `elevateErrors` function is needed. This line looks like it could just be changed to: `errors.flat(5).filter(e => !!e);`, swapping out the `5` for a constant for max nested depth or something. I may be wrong as haven't used `flat` that much.

LewisClark-HO (commented):
One comment but otherwise looks like a simple solution to an ugly problem :)

src/utils/Validate/validateContainer.js
charles474: A valid point, the reason why I haven’t used the flat feature is that it will be very limited to specified number of nesting where as if we were to then add another later of nesting, we’d also have to remember to increase that limit if that makes sense. I took the path of letting it recursively loop through the errors array.

LewisClark-HO (approved):
LGTM

PR #272, opened 2022-10-26
EAB-1176 Improve empty collection CYA handling

gardnersam:
This PR fixes a bug where `show_when` configuration wasn't being applied for components within a page collection at the CYA page. With this change `show_when` will be applied at CYA render.


LewisClark-HO (approved):
Looks good.

PR #278, opened 2022-10-31
BFCMS-361: Fix nested use component overrides

paulfgraham:
### Description
In a form page, you can set up a `"use"` property, which will find a specified component in the top-level components or, as was introduced recently, one nested within the top-level components.

It's also possible to subsequently override some of the properties from that top-level component when using it on a page. However, doing so for nested components didn't work correctly and this commit fixes that issue.

https://jira.bics-collaboration.homeoffice.gov.uk/browse/BFCMS-361

### To test
The original functionality is covered by most of the existing unit tests around `useComponent` but the one for testing the use of nested components was set up incorrectly, so fixing that in amongst this.


charles474 (approved):
LGTM

PR #279, opened 2022-10-31
COP-12196 - Fixes full_path within components in containers in a collection

charles474:
# Description
When viewing the CYA page, the `full_path` property of a component within a container may not be set properly which leads to a scenario where on the CYA page, the component has a value however it is not rendered.


Deepak-Bagrawat (approved):
LGTM

PR #280, opened 2022-11-01
Cop 9277/lm email

fisherrj2:
Bug fix to support Manage My Profile.  Need to convert ref data ID to string before comparison, and wording change to onPageAction header in FormRenderer.  To support ticket COP-9277.


aliktb (approved):
LGTM

PR #293, opened 2022-11-16
WIP: Nested component values

charles474:
Fixes a bug where nested component(s) values would not be set correctly within a container/ collections.


src/utils/Component/getComponent.js
```diff
@@ -192,8 +192,16 @@ const getComponentByType = (config) => {
  * @param {*} child the child configuration
  */
 const getChildJsx = (parent, child) => {
+  if (parent.full_path) {
+    const pathParts = parent.full_path.split('.');
+    pathParts.pop();
+    pathParts.push(child.fieldId);
+    const full_path = pathParts.join('.');
+    child.full_path = full_path;
```
LewisClark-HO, line 200: Think line 199 could be removed and just do the assignment here.

src/utils/Component/getComponent.js
```diff
@@ -192,8 +192,16 @@ const getComponentByType = (config) => {
...
+  }
+
   if (parent.formData) {
-    child.value = parent?.formData?.[child.fieldId] ? parent.formData[child.fieldId] : '';
+    child.value = parent.formData[child.fieldId] || Data.getSource(parent.formData, child.full_path);
```
LewisClark-HO, line 204: I think it would be best to set `child.value` to explicitly be `''` in the case that both of these values are `undefined`. Just because that's how it was before and don't want to break anything.

charles474:
CLOSING DUE TO BUILD FAILURE. WILL RAISE ANOTHER.

PR #295, opened 2022-11-17
Fix broken show_when for containers within partial collections

gardnersam:
`Show_whens` on containers within partial collection pages were not working as `showComponentCYA` still applies the `show_when` criteria for the container as a whole when dealing with the inner components. At that point the data for the collection is no longer available as on like 12 the passed in form data is reduced to just that relevant for the containers inner components. 

This fixes that by offering the formData input at each check which will contain the current partial-collection item.


Deepak-Bagrawat (approved):
LGTM

PR #296, opened 2022-11-22
Add ability to hide unvisited page-collection pages from the CYA screen

LewisClark-HO:
### Description
This PR adds the ability to hide certain pages in a page-collection from the CYA screen by specifying a `route` in their `collection` field.

### Examples
The storybook has been updated with an example of the new functionality.

### Testing
New unit tests have been added and existing tests expanded to cover the new functionality.


src/utils/CheckYourAnswers/getCYARowsForCollectionPage.js
```diff
@@ -3,30 +3,22 @@ import { Utils } from '@ukhomeoffice/cop-react-components';
 
 // Local imports
 import elevateNestedComponents from '../Component/elevateNestedComponents';
+import meetsCondition from '../Condition/meetsCondition';
+import getSourceData from '../Data/getSourceData';
```
gardnersam, line 7: few github warnings around here

src/utils/CheckYourAnswers/getCYARowsForCollectionPage.js
```diff
@@ -3,30 +3,22 @@ import { Utils } from '@ukhomeoffice/cop-react-components';
...
     id: item.id,
     fieldId: item.id,
     type: ComponentTypes.CONTAINER,
     required: true,
     full_path,
-    components: allComponents.map(component => {
-      if(showComponentCYA(component, item)){
-        return {
-          ...component,
-          label: Utils.interpolateString(component.label, { ...item, index: labelCount }),
```
gardnersam, line 24: does this interpolation happen somewhere else now

src/utils/CollectionPage/mergeCollectionPages.js
```diff
@@ -1,15 +1,32 @@
...
-    title: page.collection.name.charAt(0).toUpperCase() + page.collection.name.slice(1) 
+    title: page.collection.name.charAt(0).toUpperCase() + page.collection.name.slice(1),
+    components: getComponentsWithRouting(page)
   };
 };
 
 const mergeIntoMasterPage = (page, masterPage) => {
   masterPage.collection = { ...masterPage.collection, ...page.collection };
-  masterPage.components = [ ...masterPage.components, ...page.components ];
+  masterPage.components = [ ...masterPage.components, ...getComponentsWithRouting(page) ];
```
gardnersam, line 29: masterPage components already has the results from getComponentsWithRouting(), from when it's called on 23,  this looks like it would just add the same components on again

src/docs/PageCollectionExamples.stories.mdx
```diff
@@ -192,3 +192,47 @@ to the start of one if they select `Yes`, or on to the CYA screen if they select
 }
 ```
 
+## `Hiding unused routes through a collection from the CYA screen`
```
gardnersam, line 195: Would be good to explain why they can't just use a show_when here. 

My understanding is that it's because the show_when at the form level would end up with the pages / components getting rejected earlier on

src/utils/CollectionPage/mergeCollectionPages.js
LewisClark-HO: `createMasterPage` will only be called for the first page in a collection, `mergeIntoMasterPage` is called on any others. So both functions will never be called on the same page

src/utils/CheckYourAnswers/getCYARowsForCollectionPage.js
LewisClark-HO: I think it does, following it through it looks like it happens on line 29 of `getCYARow.js` - that `key` property is what's used as the label on the CYA screen.

LewisClark-HO: I've just tested it and interpolated labels are still correctly appearing on the CYA. Good spot though, didn't think to check this

PR #305, opened 2022-11-30
Cop 12161/rework tasklist flow (CYSA)

bowerscodes-COP:
Rework TaskList flow so that a 'Check your saved answers' screen is displayed when resuming a saved form, per [Jira Ticket COP-12161](https://support.cop.homeoffice.gov.uk/browse/COP-12161).


src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -60,7 +62,17 @@ const CheckYourAnswers = ({
...
+      const indexOfCurrentPage = pagesWithRows.map(object => object.id).indexOf(currentPage);
+      if (indexOfCurrentPage === 0) {
+        setPages(pagesWithRows.slice(0, 1));
+      } else {
+        setPages(pagesWithRows.slice(0, (indexOfCurrentPage)));
+      }
+    } else {
+      setPages(pagesWithRows);
+    };
   }, [_pages, onRowAction, setPages]);
```
fisherrj2, line 76: Looking at the GitHub Actions you will need to update the dependency array.  This in turn could affect behaviour so be careful to test again.

src/context/HooksContext/HooksContext.jsx
```diff
@@ -6,7 +6,7 @@ const DEFAULT_HOOKS = {
   onFormComplete: () => {},
   onFormLoad: () => {},
   onGetComponent: (config, wrap) => null,
-  onPageChange: (pageId) => pageId,
+  onPageChange: (pageId, state) => pageId,
```
fisherrj2, line 9: Just wondered if you think it is necessary to add state here?  I wasn't sure how these default hooks are used and whether it was strictly necessary to add state as shown here.

src/components/CheckYourAnswers/CheckYourAnswers.jsx
```diff
@@ -105,7 +117,7 @@ const CheckYourAnswers = ({
   return (
     <div className={DEFAULT_CLASS}>
       {title && !hide_title && (
-        <LargeHeading key='heading'>{title}</LargeHeading>
+        <LargeHeading key='heading'>{cysa ? 'Check your saved answers' : title}</LargeHeading>
```
fisherrj2, line 120: Is it worth adding this to a constant or even to the translation.json?

src/components/FormRenderer/FormRenderer.jsx
```diff
@@ -185,15 +195,15 @@ const InternalFormRenderer = ({
     }
   }, [currentTask.fullPages, data, hubDetails?.sections, hubDetails?.nonSequential]);
 
-  const onPageChange = (newPageId) => {
+  const onPageChange = (newPageId, state) => {
     clearErrors();
     setPageId(newPageId);
-    hooks.onPageChange(newPageId);
+    hooks.onPageChange(newPageId, typeof state === 'undefined' ? {} : state);
```
fisherrj2, line 201: Is this check the best approach or should we ensure that state is always populated?

src/components/CheckYourAnswers/CheckYourAnswers.jsx
bowerscodes-COP: Added as constant

fisherrj2 (approved):
Approved on the understanding that testing will be updated to ensure new features are covered.

PR #309, opened 2022-12-06
COP-12096/Checkbox Spacing & Accessibility fixes

bowerscodes-COP:
Update components version to include recent accessibility fixes


aliktb (approved):
Looks OK to me

aliktb (changes requested):
You need a new form renderer version in `package.json`

aliktb (approved):
approved

PR #311, opened 2022-12-09
COP-12543: validate email in lowercase

fisherrj2:
Fix for bug where user is able to set own email address in line manager or delegate change request in Manage My Profile. Validation was not happening if mixed case was used in the email address.
Fixes both of the following:
https://support.cop.homeoffice.gov.uk/browse/COP-12543
User is able to set their own email address as delegate
https://support.cop.homeoffice.gov.uk/browse/COP-12544
User is able to set their own email address as Line Manager


bowerscodes-COP (approved):
Looks good

PR #312, opened 2022-12-09
Cop 12094/page title

fisherrj2:
Fix for bug COP-12094 Page title on error should be 'Error:' plus the original page title.


aliktb (approved):
Looks good to me

PR #320, opened 2022-12-15
EAB Add or type show whens to task routes

gardnersam:
The show_when logic for the 'or' type does not expect an array of conditions, but an object with the array of conditions nested under a 'conditions' array. This change checks if that is the type of show_when being added and handles it appropriately.


src/utils/Component/addShowWhen.js
```diff
@@ -22,7 +22,12 @@ const addShowWhen = (component, condition) => {
       [ ...result.show_when, { ...condition } ] :
       [ result.show_when, { ...condition  } ];
   } else {
-    result.show_when = [{ ...condition }];
+    if(condition.type === 'or'){
```
LewisClark-HO, line 25: Taking a leaf from Mark's book - maybe this bit could be changed to a ternary to make it a little more concise? 
```
result.show_when = 
  condition.type === 'or' ?
  { ...condition } :
  [{ ...condition }];
```

LewisClark-HO (approved):

PR #321, opened 2022-12-15
EAB-996 Allow disabling validation for CYA saveandcontinue actions

gardnersam:
The aim of this PR is to allow using a custom Save and Continue type action on a CYA page where validation is set to false. 

There is currently code which defaults the validation to true which I don't think is needed or helpful so is removed. Code to check the validation flag is then added to the CYA Save and Continue action code.


charles474 (approved):
LGTM

PR #323, opened 2022-12-16
Fix update of setHubDetails, solving render and console error

mark-roberts-ho:
Solves an issue causing TaskList to display error messages and failure to render


gardnersam (approved):
Looks good, I especially like the condensing of the tests in TaskLIst.test.js - much more readable!

PR #326, opened 2022-12-20
COP-12094: reset document title before error check

fisherrj2:
When an error is raised the document title should display 'Error: ' before the original document title.  This fix is to switch from using a state variable which has an unusual side effect, and instead to reset the document title to remove error prefix in case it is there before the error check is done. See COP-12094 for ticket details.


keenan218 (approved):
Looks good

PR #327, opened 2022-12-22
Specify class depending on tag for html

fisherrj2:
For an HTML component `className` was missing from the call to `Markup` so that a default was used.  New components have been added for `paragraph` and `list` (ordered and unordered) that override `html` and specify tag-specific `className`. See ticket COP-12153 for details.


aliktb (commented):
Requires bump on version in `package.json`

src/utils/Component/getComponent.js
```diff
@@ -92,7 +92,20 @@ const getHeading = (config) => {
 
 const getHTML = (config) => {
   const attrs = cleanAttributes(config);
-  return <Markup {...attrs} />;
+  let className;
+  switch (attrs.tagName) {
+    case 'p':
+      className = "govuk-body";
```
paulfgraham, line 98: This will change the markup of every single team that's using the Form Renderer and has a `<p>` tag anywhere on their form, which will be pretty much everyone.

src/utils/Component/getComponent.js
```diff
@@ -92,7 +92,20 @@ const getHeading = (config) => {
...
+      break;
+    case 'ul':
+      className = "govuk-list govuk-list--bullet";
+      break;
+    case 'ol':
+      className = "govuk-list govuk-list--number";
+      break;
+    default:
+  }
+  return <Markup {...attrs} className={className} />;
```
paulfgraham, line 108: The `config` parameter already supports a `className` property so if anyone is using it (I think we are in some forms), this will override whatever has been specified. Can this be set up to either use what's specified or merge the one above with the one that's been specified?

However, as mentioned, defaulting or forcing it here is technically a breaking change anyway.

src/utils/Component/getComponent.js
paulfgraham, line 98: All of these values should be single quotes.

paulfgraham (changes requested):

paulfgraham:
There's no version number change on this PR.

src/utils/Component/getComponent.js
fisherrj2: Changed to consistently use single quotes

fisherrj2: But is there a scenario where the specified class would not be used for the `<p>` tag?  Also note that the font size is defaulted in the Markup component.

src/utils/Component/getComponent.js
fisherrj2: Yes that can be done but if this is a breaking change then I am wondering how to meet the requirement without breaking the function.

paulfgraham: My proposal would be to have a specific component type of `"p"` or `"paragraph"` that uses the `<Markup />` component behind the scenes:

```json
"components": [
  {
    "type": "html",
    "tagName": "p",
    "content": "Hello world"
  }
]
```
This would output simply `<p>Hello world</p>`, as it currently does. Therefore not a breaking change.

```json
"components": [
  {
    "type": "p",
    "content": "Hello world"
  }
]
```
This would have a `case` switch in in `utils/Component/getComponent` for `ComponentTypes.PARAGRAPH` (also newly-added) that calls this:
```js
const getParagraph = (config) => {
  const attrs = cleanAttributes(config);
  return getHTML({
    className: 'govuk-body',
    ...attrs,
    tagName: 'p',
    type: ComponentTypes.HTML
  };
};
```
And this would then output `<p className="govuk-body">Hello world</p>`.

Doing the attributes in that order would also allow you to override the `className` in the `config` if you _really_ wanted to - as in, you could set it to, say, `govuk-body highlight` or some such thing. If you don't want to support that, simply put the `className: 'govuk-body'` after the `...attrs`.

This would be a minor release to reflect the new paragraph type being supported and all existing forms are complete unaffected.

src/utils/Component/getComponent.js
paulfgraham: Irrespective of whether anyone is _currently_ overriding the `className`, there are a bunch of forms in the `cop-forms` repository that have paragraphs (I haven't checked for the overriding) and this _will_ impact them.

The thing to understand here is that, whilst it's probably a good thing overall to set the class on standard paragraphs in the forms to `"govuk-body"`, it's not _currently_ being done, which means all the forms currently out there in the wild don't have that styling for their paragraphs. Suddenly introducing it will visually impact all of those paragraphs.

Also, on the subject of the "default" font size in the markup, that shouldn't be there! And that change was _precisely_ the same sort of change that I'm talking about with this one - when it got added (and _forced_ to be 19px, which isn't even responsive!), it broke a whole bunch of forms. You'd be similarly breaking a bunch of forms by _forcing_ them to have the `govuk-body` class.

fisherrj2: Thanks for your comments - it is much appreciated and will prompt a rethink.

src/utils/Component/getComponent.js
fisherrj2: Do you see the same approach being used for `<ul>` and `<ol>` tags is that not necessary?

paulfgraham:
@fisherrj2, @aliktb, @willrichards:

This library is currently being used by at least 3 applications (that I know of) and other HO teams with other React applications are lining up to use it too. It's really important to _not_ make breaking changes in here unless they _really_ need to be made... and when they _do_ need to be made, it has to go in as a major version bump so those consuming applications can make an active choice to go up a version rather than their `npm install` grabbing the latest "compatible" version and suddenly finding it's broken something.

This library and the component library that it imports are _not_ on `0.x.y` and are both being used in production environments. It's critical to do proper semantic versioning and to understand the scope and scale of the impact of any and all changes in here.

* If you fix a genuine _bug_ (and this one _isn't_ a bug, whatever the description or Jira ticket says),_ that's a **PATCH** release.
* If you add some new functionality (as outlined above with support for a specific paragraph type), that's a **MINOR** release.
* If you change the attributes, change existing behaviour, change the existing appearance (this includes the `className`), or change any of the defaults for those attributes, that's a **MAJOR** release because it impacts _everyone_.

I'm not blaming anyone I've tagged in this comment or singling out any contributors to this repository (or the component library one) but we've had _many_ instances of breaking changes being introduced and it adversely impacting the consuming applications. This _has_ to stop happening.

_Wherever possible_, we should avoid doing a major release because that means those using the older major release are forced to adopt the newer major version and deal with its breaking changes if they want any of the newer stuff or bug fixes. That's just plain unfair and not good practice.

src/utils/Component/getComponent.js
paulfgraham: Lists are a problem as it currently stands because they don't render very nicely anyway. I'd therefore be inclined to do them as distinct types and get away from the `"content"` string. Something like this would work, I'd suggest (not written in an IDE - just in here - so no guarantees it'll work):
```json
"components": [
  {
    "type": "list",
    "ordered": true,
    "items": [
      "item 1",
      "item 2",
      "Or maybe some <span>markup</span> in here"
    ]
  }
]
```

```js
const getList = (config) => {
  const attrs = cleanAttributes(config);
  const tagName = attrs.ordered ? 'ol' : 'ul';
  const bullet = attrs.ordered ? 'number' : 'bullet';
  const className = `govuk-list govuk-list--${bullet}`;
  const content = attrs.items?.map((item) => {
    return <li>{item}</li>;
  }).join('');
  return getHTML({
    className,
    ...attrs,
    tagName,
    content,
    type: ComponentTypes.HTML
  };
};
```

src/utils/Component/getComponent.js
```diff
@@ -95,6 +95,33 @@ const getHTML = (config) => {
   return <Markup {...attrs} />;
 };
 
+const getParagraph = (config) => {
```
paulfgraham, line 98: _Nitpick:_
It'd be good to keep these in alphabetical order.

In fact, not for this PR, but it's probably at a point now where it'd be sensible to move these functions into their own files.

src/utils/Component/getComponent.js
```diff
@@ -95,6 +95,33 @@ const getHTML = (config) => {
...
+  const attrs = cleanAttributes(config);
+  return getHTML({
+    className: 'govuk-body',
```
paulfgraham, line 101: It'd probably be good to make this a constant rather than a string literal here.

src/utils/Component/getComponent.js
```diff
@@ -95,6 +95,33 @@ const getHTML = (config) => {
...
+    tagName: 'p',
+    type: ComponentTypes.HTML
+  });
+};
+
+const getList = (config) => {
+  const attrs = cleanAttributes(config);
+  const tagName = attrs.ordered ? 'ol' : 'ul';
+  const bullet = attrs.ordered ? 'number' : 'bullet';
+  const className = `govuk-list govuk-list--${bullet}`;
```
paulfgraham, line 112: Same thing as above with making `'govuk-list'` a constant... and then you can do this for `className` (assuming your constant is called `LIST_CLASS`):

```js
const classes = Utils.classBuilder(LIST_CLASS, bullet);
```

`Utils` is imported from the `src/utils` directory.

paulfgraham: The benefit of using `Utils.classBuilder` is that it already handles the string literal stuff. Actually, you could theoretically also pass in `config.className` as the third parameter to pick up anything specified in the JSON... and then swap lines 117 and 118, below.

src/utils/Component/getComponentTests/getComponent.paragraph.test.js
```diff
@@ -0,0 +1,22 @@
...
+    const CONTENT = 'paragraph content';
+    const COMPONENT = { type: ComponentTypes.PARAGRAPH, content: CONTENT, 'data-testid': ID };
+    const { container } = render(getComponent(COMPONENT));
+
+    const p = getByTestId(container, ID);
+    expect(p.innerHTML).toContain(CONTENT);
+    expect(p.tagName).toEqual('P');
+    expect(p.className).toEqual('govuk-body');
+  });
+
```
paulfgraham, line 21: This would benefit from a second test to make sure specifying `className` in the JSON overrides the `govuk-body`, which is how you've set up the function to behave.

src/utils/Component/getComponent.js
fisherrj2: Changed to be in alphabetical order

src/utils/Component/getComponent.js
fisherrj2: now a constant

src/utils/Component/getComponent.js
fisherrj2: uses constant and class builder

src/utils/Component/getComponentTests/getComponent.paragraph.test.js
fisherrj2: added test for className override on paragraph component, but prevented override on list component as no requirement exists

paulfgraham (approved):

PR #332, opened 2023-01-05
EAB-1331 remove unwanted nested data in collections

gardnersam:
When a user enters data in a component that is nested under a radio button option, then selects another radio option and enters data there, we should not keep the data from the first nested component. This was not a problem generally as `cleanHiddenNestedData` handled it on page submission. However this did not work if the page the top level component was on was within a collection as only top level components were looked at. This PR fixes that bug by also scanning through components within the collection the page is a part of.


LewisClark-HO (approved):
LGTM

